Issue #126, opened 2014-04-09
Cursor reuse


benbjohnson:
Cursors should be allowed to be reused so that iterating over a multi-level bucket structure only requires one cursor for each level.

Some kind of function like this would be required:

``` go
func (b *Bucket) InitCursor(*Cursor)
```

That's not quite the right name though.

Issue #125, opened 2014-04-09
Optimize nested buckets for read-only transactions


benbjohnson:
Buckets are currently tracked at every level for all transactions. This should only be required for writable transactions.

Issue #124, opened 2014-04-09
Inline Buckets


benbjohnson:
Allow small nested buckets to have their data written inline with their bucket header instead of requiring an entire root page.

PR #127, opened 2014-04-09
Add nested buckets.

### LINKED ISSUE #56: Nested Keys
## Overview

Add the ability to nest key/value data inside of keys. This is useful for building indexes and for certain types of nested data.
## API

We'll need a function for retrieving a cursor to iterate over the nested values:

```
func (c *Cursor).Subcursor(key string) (*Cursor, error)
```

We'll also need functions to get/put/delete. These subkeys are essentially the same thing as buckets. Maybe we just return a `Bucket`:

```
func (b *Bucket) Sub(key string) *Bucket
```

Otherwise we'll have to create a bunch of methods on Bucket:

```
func (b *Bucket) SubGet(key []byte) []byte
func (b *Bucket) SubPut(key []byte, []byte) 
func (b *Bucket) SubDelete(key []byte, []byte) 
```

That's pretty ugly. I'm leaning toward just returning a sub-bucket.
## Changes
- [ ] Add a flag on the leaf node element to distinguish it as having nested data.
- [ ] Add `Subcursor()` function to the `Cursor()`.
- [ ] Delete subpages for when deleting the parent key.
## Caveats

Only support single-level subcursors. Not only will it make it less confusing but it means that Cursor's can reuse a single subcursor during iteration.

Subbuckets introduce a new error condition where you could try to do a normal `Bucket.Get()` on a nested key. Currently `Get()` doesn't return an error which I'd like to keep. I will probably just have it return `nil`. The other option is to panic.

benbjohnson:
## Overview

This pull request adds the ability to nest buckets inside of other buckets. It also replaces the buckets page with a root bucket. This pull request looks bigger than it is. It's mostly moving a lot of `Tx` functionality over to the `Bucket`.
## API Changes

Previously, the ability to create, retrieve, and delete buckets was entirely in `Tx` and `buckets`. This made sense because buckets were a top-level construct. Now that buckets can exist inside other buckets, all this functionality has moved into the `Bucket` type.

Buckets and simple byte slice values all exist in the same keyspace. However, trying to create a bucket named `foo` and simple value with the `foo` key will return an `ErrIncompatibleValue` error.
### New Bucket functions:

``` go
func (b *Bucket) Bucket(name []byte) *Bucket
func (b *Bucket) CreateBucket(key []byte) error
func (b *Bucket) CreateBucketIfNotExists(key []byte) error
func (b *Bucket) DeleteBucket(key []byte) error
```
### Changes in Bucket functions:
- `Bucket.Get()` returns `nil` when a bucket key is provided.
- `Bucket.Put()` returns `ErrIncompatibleValue` when putting to an existing bucket key.
- `Bucket.Delete()` returns `ErrIncompatibleValue` when deleting an existing bucket key.
### Tx functions

All the previous Tx bucket functions still exist and simply wrap the root bucket's same functions:

``` go
func (tx *Tx) Bucket(name []byte) *Bucket
func (tx *Tx) CreateBucket(name []byte) error
func (tx *Tx) CreateBucketIfNotExists(name []byte) error
func (tx *Tx) DeleteBucket(name []byte) error
```

However, since buckets are in the regular keyspace, their names are byte slices instead of strings.
## File Format Changes

Allowing buckets to be nested made the `buckets` page redundant. Now, instead of a specialized page type, there is a root bucket that the `Tx` uses to find and create top-level buckets. This root bucket is only accessible to the `Tx` so it only stores buckets (e.g. no simple byte slice values).

Because of this change, older Bolt databases are not compatible after this pull request is merged. I'll be merging in an import/export tool through a separate PR before this pull request is merged. That will allow databases to be exported to a generalized JSON format and imported into a new version of Bolt.

The data format `version` has been bumped to `2` with this PR. Opening previous data format versions of the database will return `ErrVersionMismatch` error.
## Upgrading

To upgrade a Bolt database that uses the previous version (`1`), you'll need to install Bolt from the `data/v1` tag:

``` sh
$ cd $GOPATH/src/github.com/boltdb/bolt
$ git fetch origin
$ git checkout "data/v1"
$ go install ./cmd/bolt
$ bolt export path/to/db > dump.json
```

Then install Bolt from `master`:

``` sh
$ git checkout master
$ go install ./cmd/bolt
$ bolt import --input dump.json path/to/newdb
```

Now you can swap out your old database:

``` sh
$ mv path/to/db path/to/db.bak
$ mv path/to/newdb path/to/db
```
## Other Notes

This pull request lacks a few optimizations that will be coming in future PRs:
1. Inline buckets (#124) - Currently buckets require their own root page. Inlining buckets would allow small buckets (`< pageSize`) to be written with their data into the same value that their `bucket` header exists.
2. Allocations (#125) - Currently subbuckets require allocation of a `Bucket` for access. This is mostly to track nodes for each bucket but that shouldn't be required for read-only transactions. Tweaking this should make iteration faster for nested buckets.
3. Cursor Reuse (#126) - If someone is iterating over two levels of buckets then they should be able to only allocate two cursors and reuse them for each level. Currently Bolt requires a new cursor after every change in a higher level bucket.

---

Fixes #56.

/cc @tv42 @snormore @mkobetic

### DIFF
--- a/cmd/bolt/buckets_test.go
+++ b/cmd/bolt/buckets_test.go
@@ -13,9 +13,9 @@ func TestBuckets(t *testing.T) {
 	SetTestMode(true)
 	open(func(db *bolt.DB, path string) {
 		db.Update(func(tx *bolt.Tx) error {
-			tx.CreateBucket("woojits")
-			tx.CreateBucket("widgets")
-			tx.CreateBucket("whatchits")
+			tx.CreateBucket([]byte("woojits"))
+			tx.CreateBucket([]byte("widgets"))
+			tx.CreateBucket([]byte("whatchits"))
 			return nil
 		})
 		db.Close()
--- a/cmd/bolt/export.go
+++ b/cmd/bolt/export.go
@@ -23,25 +23,33 @@ func Export(path string) {
 	}
 	defer db.Close()
 
-	db.View(func(tx *bolt.Tx) error {
+	err = db.View(func(tx *bolt.Tx) error {
 		// Loop over every bucket and export it as a raw message.
 		var root []*rawMessage
-		for _, b := range tx.Buckets() {
+		err := tx.ForEach(func(name []byte, b *bolt.Bucket) error {
 			message, err := exportBucket(b)
 			if err != nil {
 				fatal(err)
 			}
+			message.Key = name
 			root = append(root, message)
+			return nil
+		})
+		if err != nil {
+			return err
 		}
 
 		// Encode all buckets into JSON.
 		output, err := json.Marshal(root)
 		if err != nil {
-			fatal("encode: ", err)
+			return fmt.Errorf("encode: ", err)
 		}
 		print(string(output))
 		return nil
 	})
+	if err != nil {
+		fatal(err)
+	}
 }
 
 func exportBucket(b *bolt.Bucket) (*rawMessage, error) {
@@ -50,11 +58,22 @@ func exportBucket(b *bolt.Bucket) (*rawMessage, error) {
 	err := b.ForEach(func(k, v []byte) error {
 		var err error
 
+		// If there is no value then it is a bucket.
+		if v == nil {
+			child, err := exportBucket(b.Bucket(k))
+			if err != nil {
+				return fmt.Errorf("bucket: %s", err)
+			}
+			child.Key = k
+			children = append(children, child)
+			return nil
+		}
+
+		// Otherwise it's a regular key.
 		var child = &rawMessage{Key: k}
 		if child.Value, err = json.Marshal(v); err != nil {
 			return fmt.Errorf("value: %s", err)
 		}
-
 		children = append(children, child)
 		return nil
 	})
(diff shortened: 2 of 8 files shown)


ericnp:
I don't see in the docs or in the tests how to traverse the bucket trees (i.e. nested buckets). It seems when cursor or ForEach hits upon a sub-bucket key it returns the value as nil. Is that the only case where you can get nil; i.e., if you see nil value in that context is that guaranteed to be a subbucket? And then how do you access the sub-bucket to traverse *it*? I guess it's actually stored at the root level of the db? Meaning all buckets actually share that one namespace and you have to worry about naming collisions among all subbuckets and top level buckets? Sorry if this is a stupid question. I'd be happy to work on embellishing the documentation with this stuff if you welcome contributions like that.

Awesome software btw, thank you for open sourcing :)

tv42:
http://godoc.org/github.com/boltdb/bolt#Cursor
"Cursors see nested buckets with value == nil."

http://godoc.org/github.com/boltdb/bolt#Bucket.Bucket
"Bucket retrieves a nested bucket by name."

ericnp:
Oh, right, Bucket.Bucket. Thanks.

Issue #257, opened 2014-09-18
Does not compile or work properly on OpenBSD


davecgh:
As the title indicates, bolt does not compile on OpenBSD.  I have done a bit of preliminary work on it as I will detail below to make it compile, but then it fails the tests.

First, the compile errors are as follows:

> ./db.go:107: undefined: flock
> ./db.go:180: undefined: mmap
> ./db.go:201: undefined: munmap
> ./db.go:266: undefined: fdatasync
> ./db.go:298: undefined: funlock
> ./tx.go:263: undefined: odirect
> ./tx.go:429: undefined: fdatasync
> ./tx.go:452: undefined: fdatasync

The reason for this is the build tags in `bolt_unix.go` don't include openbsd (or freebsd/netbsd for that matter).  Also, since there is no bolt_openbsd.go, simply adding the build tag results in missing definitions.

So, to allow it to compile, I copied `bolt_darwin.go` to `bolt_openbsd.go` (darwin is based on BSD) and added `openbsd` to the build tags in `bolt_unix.go`.

``` diff
diff --git a/bolt_openbsd.go b/bolt_openbsd.go
new file mode 100644
index 0000000..6528f72
--- /dev/null
+++ b/bolt_openbsd.go
@@ -0,0 +1,12 @@
+package bolt
+
+import (
+       "os"
+)
+
+var odirect int
+
+// fdatasync flushes written data to a file descriptor.
+func fdatasync(f *os.File) error {
+       return f.Sync()
+}
diff --git a/bolt_unix.go b/bolt_unix.go
index f7d2fe5..59de101 100644
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -1,4 +1,4 @@
-// +build linux darwin
+// +build linux darwin openbsd
```

This allows it to properly compile, but running the tests now results in:

``` bash
$ go test
seed: 63588
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000


consistency check failed (1 errors)
page 3: reachable freed

db saved to:
/tmp/bolt-110493209


exit status 255
FAIL    github.com/boltdb/bolt       0.064s
```

PR #258, opened 2014-09-18
Rework build to support more liberal unix variants.

davecgh:
Rather than having the build setup such that it will only work on the
specifically defined operating systems, this commit modifies it to use
more liberal !windows,!plan9 build tag for the unix specific bits.

This means bolt will compile on more Operating Systems such as OpenBSD,
FreeBSD, and NetBSD.

See boltdb/bolt#257 for discussion.

I've tested these changes with cross compilation to windows, darwin, linux, openbsd, and freebsd.

It should also work fine for NetBSD and DragonFly BSD, but I did not test those.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -1,4 +1,4 @@
-// +build linux darwin
+// +build !windows,!plan9
 
 package bolt
 
--- a/boltsync_unix.go
+++ b/boltsync_unix.go
@@ -1,3 +1,5 @@
+// +build !windows,!plan9,!linux
+
 package bolt
 
 import (


cdorbell:
Hi
i evaluate boltdb on my freebsd via the install of influxd , the compilation is working but i can't create a new database via influx client.
i tested directly the compilation it's working but the go test failed
 % go test
seed: -970
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000
runtime: garbage collector found invalid heap pointer *(0x826b6a0+0x4c)=0x583a9000 s=nil
fatal error: invalid heap pointer
[...]
goroutine 4 [finalizer wait]:
runtime.gopark(0x806f780, 0x82715b4, 0x81b41c8, 0xe)
        /usr/local/go/src/runtime/proc.go:130 +0xd8 fp=0x3836ef98 sp=0x3836ef80
runtime.goparkunlock(0x82715b4, 0x81b41c8, 0xe)
        /usr/local/go/src/runtime/proc.go:136 +0x3c fp=0x3836efac sp=0x3836ef98
runtime.runfinq()
        /usr/local/go/src/runtime/malloc.go:727 +0x96 fp=0x3836eff0 sp=0x3836efac
runtime.goexit()
        /usr/local/go/src/runtime/asm_386.s:2287 +0x1 fp=0x3836eff4 sp=0x3836eff0
created by runtime.createfing
        /usr/local/go/src/runtime/malloc.go:707 +0x53
exit status 2
FAIL    github.com/boltdb/bolt  12.127s

my system is under freebsd 10.0 amd64

it is the same fixe that you use under openbsd ?

Rgds

Issue #277, opened 2015-01-07
panic: index out of range


tgulacsi:
Hi,

When using BoltDB as an index DB for Camlistore, I got a panic, and an unusable boltdb:
http://www.gthomas.eu/gthomas/camli-index.boltdb.index_out_of_range

```
2015/01/06 12:42:52 http: panic serving 127.0.0.1:48279: runtime error: index out of range
goroutine 681 [running]:
net/http.func·011()
        /usr/local/go/src/net/http/server.go:1130 +0xa8
camlistore.org/third_party/github.com/boltdb/bolt.(*Tx).Commit(0x26a5bb00, 0x0, 0x0)
        /home/gthomas/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/third_party/github.com/boltdb/bolt/tx.go:161 +0x2dd
camlistore.org/pkg/sorted/boltdb.(*kvis).CommitBatch(0x191f6c00, 0xb7566518, 0x21aceb30, 0x0, 0x0)
        /home/gthomas/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/sorted/boltdb/boltdb.go:174 +0x122
camlistore.org/pkg/index.(*Index).commit(0x19240f00, 0x21ace9e0, 0x0, 0x0)
        /home/gthomas/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/index/receive.go:272 +0x1ab
camlistore.org/pkg/index.(*Index).ReceiveBlob(0x19240f00, 0xb7563348, 0x21d7f0e0, 0xb75662b0, 0x21d7f100, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/gthomas/src/camlistore.org/tmp/build-gopath-nosqlite/src/camlistore.org/pkg/index/receive.go:239 +0x673
camlistore.org/pkg/serverinit.(*storageAndConfig).ReceiveBlob(0x1923dab0, 0xb7563348, 0x21d7f0e0, 0xb75662b0, 0x21d7f100, 0x0, 0x0, 0x0, 0x0, 0x0)
        <autogenerated>:56 +0x8e
camlistore.org/pkg/blobserver.receive(0xb7566298, 0x1923dab0, 0xb7563348, 0x21d7f0e0, 0xb75662b0, 0x21d7f100, 0x82cc001, 0x0, 0x0, 0x0, ...)
[...]

PR #281, opened 2015-01-12
Raise maxMapSize from 256MB to 2GB on 386, arm.

### LINKED ISSUE #260: Cursor behavor on modifications
Hello.

https://gist.github.com/freepk/b4bfb0aacbc6e2a5f862

There are many situations when cursor's Prev() Next() don't return correct position or error after data change.
In this example cursor position after Seek is "testKey00". After delete "testKey00" and "testKey01", Next() return nil, but Prev() return "testKey02" how to understan this behavior?

tgulacsi:
Fixes #260 and #277.

### DIFF
--- a/bolt_386.go
+++ b/bolt_386.go
@@ -1,4 +1,4 @@
 package bolt
 
 // maxMapSize represents the largest mmap size supported by Bolt.
-const maxMapSize = 0xFFFFFFF // 256MB
+const maxMapSize = 0x7FFFFFFF // 2GB
--- a/bolt_arm.go
+++ b/bolt_arm.go
@@ -1,4 +1,4 @@
 package bolt
 
 // maxMapSize represents the largest mmap size supported by Bolt.
-const maxMapSize = 0xFFFFFFF // 256MB
+const maxMapSize = 0x7FFFFFFF // 2GB


fern4lvarez:
You mentioned that compilation failed when trying to enlarge this value. What was the issue and how did you overcome it?

tgulacsi:
I've tried the noob way: append F, then 0 to the constant.
The problem is with the high end: only 7 is allowed (2^32-1).

benbjohnson:
@tgulacsi Thanks! I added a check in the `mmap()` function to ensure Bolt doesn't try to grow beyond the max. Can you try this and see if it gives you an error (instead of a panic) when the max size is back at 256MB?

```go
// mmap opens the underlying memory-mapped file and initializes the meta references.
// minsz is the minimum size that the new mmap can be.
func (db *DB) mmap(minsz int) error {
	db.mmaplock.Lock()
	defer db.mmaplock.Unlock()

	info, err := db.file.Stat()
	if err != nil {
		return fmt.Errorf("mmap stat error: %s", err)
	} else if int(info.Size()) < db.pageSize*2 {
		return fmt.Errorf("file size too small")
	}

	// Ensure the size is at least the minimum size.
	var size = int(info.Size())
	if size < minsz {
		size = minsz
	}
	size = db.mmapSize(size)

	// Verify the map size is not above the maximum allowed.
	if size > maxMapSize {
		return fmt.Errorf("mmap too large")
	}

	// Dereference all mmap references before unmapping.
	if db.rwtx != nil {
		db.rwtx.root.dereference()
	}

	// Unmap existing data before continuing.
	if err := db.munmap(); err != nil {
		return err
	}

	// Memory-map the data file as a byte slice.
	if err := mmap(db, size); err != nil {
		return err
	}

	// Save references to the meta pages.
	db.meta0 = db.page(0).meta()
	db.meta1 = db.page(1).meta()

	// Validate the meta pages.
	if err := db.meta0.validate(); err != nil {
		return fmt.Errorf("meta0 error: %s", err)
	}
	if err := db.meta1.validate(); err != nil {
		return fmt.Errorf("meta1 error: %s", err)
	}

	return nil
}
```

tgulacsi:
Yes, this change prevents creating an unreadable .boltdb, and returns "mmap allocate error: mmap too large".
So LGTM.

benbjohnson:
Awesome. Thanks for checking that and thanks for bumping the max size to 2GB. I pushed the check via PR #282.

luxas:
Never mind, found the code after reading a bit more. However, we think there is an issue with `32-bit` anyway: https://github.com/coreos/etcd/pull/4715#issuecomment-194709562
If you could join the discussion it would be great.

PR #282, opened 2015-01-12
Add check for max mmap size.

benbjohnson:
The max mmap size was previous unchecked which resulted in a panic once the maximum size was reached. This commit adds a check for the max size when re-mapping and returns an error if the new map will exceed the size.

_Note: Exceeding the max size did **not** cause corruption. It simple caused a panic._

Thanks to Tamás Gulácsi (@tgulacsi) for testing out the change on i386.


benbjohnson:
Also, note that this change realistically only affects 32-bit machines as the max mmap size on 64-bit machines is 256TB.

Issue #244, opened 2014-08-24
Very low write speed in random mode.


freepk:
Hello.

<pre>
% ./bolt bench --no-sync --count '1000' --write-mode 'rnd'
# Write 2.68452ms       (2.684us/op)    (372578 op/sec)

% ./bolt bench --no-sync --count '10000' --write-mode 'rnd'
# Write 89.396821ms     (8.939us/op)    (111869 op/sec)

% ./bolt bench --no-sync --count '100000' --write-mode 'rnd'
# Write 8.09731744s     (80.973us/op)   (12349 op/sec)
</pre>

I try to run bench with --count '1000000' but have no luck with results.

Btw where is random read bench option(only sequential)?

PR #285, opened 2015-01-17
Add transaction batching

tv42:
Batcher makes it easy to make lots of small transactions with
significantly better performance. Batcher.Update combines multiple
Update calls into a single disk transaction, managing errors smartly.

I'm aware of https://github.com/boltdb/coalescer. I wrote the original version of this before you pushed coalescer, it just took a while for me to circle back to cleaning it up. This code doesn't leave a goroutine around, and is a few percent faster ;) But the real difference to coalescer is this: with this code, callers don't need to write a for loop around ErrRollback. I pull failing updaters out of the transaction, and retry the others automatically.

I'd really like to see this as part of bolt, and I feel like having a good batcher mechanism would help significantly with performance complaints like https://github.com/boltdb/bolt/issues/244 . If this looks promising, it could even sit alongside DB.Update as DB.Batch -- or even become DB.Update (though beware the "runs multiple times" behavior difference).

With this in place, I see no reason why any of my apps would ever again bother doing non-batched transactions.

Let me know what you think. If you say no, this'll live as tv42/semiauto (as opposed to bolt action..)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,48 @@ no mutating operations are allowed within a read-only transaction. You can only
 retrieve buckets, retrieve values, and copy the database within a read-only
 transaction.
 
+
+#### Batch read-write transactions
+
+Each `DB.Update()` waits for disk to commit the writes. This overhead
+can be minimized by combining multiple updates with the `DB.Batch()`
+function:
+
+```go
+err := db.Batch(func(tx *bolt.Tx) error {
+	...
+	return nil
+})
+```
+
+Concurrent Batch calls are opportunistically combined into larger
+transactions. Batch is only useful when there are multiple goroutines
+calling it.
+
+The trade-off is that `Batch` can call the given
+function multiple times, if parts of the transaction fail. The
+function must be idempotent and side effects must take effect only
+after a successful return from `DB.Batch()`.
+
+For example: don't display messages from inside the function, instead
+set variables in the enclosing scope:
+
+```go
+var id uint64
+err := db.Batch(func(tx *bolt.Tx) error {
+	// Find last key in bucket, decode as bigendian uint64, increment
+	// by one, encode back to []byte, and add new key.
+	...
+	id = newValue
+	return nil
+})
+if err != nil {
+	return ...
+}
+fmt.Println("Allocated ID %d", id)
+```
+
+
 #### Managing transactions manually
 
 The `DB.View()` and `DB.Update()` functions are wrappers around the `DB.Begin()`
--- a/batch.go
+++ b/batch.go
@@ -1,0 +1,135 @@
+package bolt
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+)
+
+// Batch calls fn as part of a batch. It behaves similar to Update,
+// except:
+//
+// 1. concurrent Batch calls can be combined into a single Bolt
+// transaction.
+//
+// 2. the function passed to Batch may be called multiple times,
+// regardless of whether it returns error or not.
+//
+// This means that Batch function side effects must be idempotent and
+// take permanent effect only after a successful return is seen in
+// caller.
+//
+// Batch is only useful when there are multiple goroutines calling it.
+func (db *DB) Batch(fn func(*Tx) error) error {
+	errCh := make(chan error, 1)
+
+	db.batchMu.Lock()
+	if (db.batch == nil) || (db.batch != nil && len(db.batch.calls) >= db.MaxBatchSize) {
+		// There is no existing batch, or the existing batch is full; start a new one.
+		db.batch = &batch{
+			db: db,
+		}
+		db.batch.timer = time.AfterFunc(db.MaxBatchDelay, db.batch.trigger)
+	}
+	db.batch.calls = append(db.batch.calls, call{fn: fn, err: errCh})
+	if len(db.batch.calls) >= db.MaxBatchSize {
+		// wake up batch, it's ready to run
+		go db.batch.trigger()
+	}
+	db.batchMu.Unlock()
+
+	err := <-errCh
+	if err == trySolo {
+		err = db.Update(fn)
+	}
+	return err
+}
+
+type call struct {
+	fn  func(*Tx) error
+	err chan<- error
+}
+
+type batch struct {
+	db    *DB
+	timer *time.Timer
+	start sync.Once
+	calls []call
+}
+
+// trigger runs the batch if it hasn't already been run.
+func (b *batch) trigger() {
+	b.start.Do(b.run)
+}
+
+// run performs the transactions in the batch and communicates results
+// back to DB.Batch.
+func (b *batch) run() {
+	b.db.batchMu.Lock()
+	b.timer.Stop()
+	// Make sure no new work is added to this batch, but don't break
+	// other batches.
+	if b.db.batch == b {
+		b.db.batch = nil
+	}
+	b.db.batchMu.Unlock()
+
+retry:
+	for len(b.calls) > 0 {
+		var failIdx = -1
+		err := b.db.Update(func(tx *Tx) error {
+			for i, c := range b.calls {
+				if err := safelyCall(c.fn, tx); err != nil {
+					failIdx = i
+					return err
+				}
+			}
+			return nil
+		})
+
+		if failIdx >= 0 {
+			// take the failing transaction out of the batch. it's
+			// safe to shorten b.calls here because db.batch no longer
+			// points to us, and we hold the mutex anyway.
+			c := b.calls[failIdx]
+			b.calls[failIdx], b.calls = b.calls[len(b.calls)-1], b.calls[:len(b.calls)-1]
+			// tell the submitter re-run it solo, continue with the rest of the batch
+			c.err <- trySolo
+			continue retry
+		}
+
+		// pass success, or bolt internal errors, to all callers
+		for _, c := range b.calls {
+			if c.err != nil {
+				c.err <- err
+			}
+		}
+		break retry
+	}
+}
+
+// trySolo is a special sentinel error value used for signaling that a
+// transaction function should be re-run. It should never be seen by
+// callers.
+var trySolo = errors.New("batch function returned an error and should be re-run solo")
+
+type panicked struct {
+	reason interface{}
+}
+
+func (p panicked) Error() string {
+	if err, ok := p.reason.(error); ok {
+		return err.Error()
+	}
+	return fmt.Sprintf("panic: %v", p.reason)
+}
+
+func safelyCall(fn func(*Tx) error, tx *Tx) (err error) {
+	defer func() {
+		if p := recover(); p != nil {
+			err = panicked{p}
+		}
+	}()
+	return fn(tx)
+}
--- a/batch_benchmark_test.go
+++ b/batch_benchmark_test.go
@@ -1,0 +1,170 @@
+package bolt_test
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"hash/fnv"
+	"sync"
+	"testing"
+
+	"github.com/boltdb/bolt"
+)
+
+func validateBatchBench(b *testing.B, db *TestDB) {
+	var rollback = errors.New("sentinel error to cause rollback")
+	validate := func(tx *bolt.Tx) error {
+		bucket := tx.Bucket([]byte("bench"))
+		h := fnv.New32a()
+		buf := make([]byte, 4)
+		for id := uint32(0); id < 1000; id++ {
+			binary.LittleEndian.PutUint32(buf, id)
+			h.Reset()
+			h.Write(buf[:])
+			k := h.Sum(nil)
+			v := bucket.Get(k)
+			if v == nil {
+				b.Errorf("not found id=%d key=%x", id, k)
+				continue
+			}
+			if g, e := v, []byte("filler"); !bytes.Equal(g, e) {
+				b.Errorf("bad value for id=%d key=%x: %s != %q", id, k, g, e)
+			}
+			if err := bucket.Delete(k); err != nil {
+				return err
+			}
+		}
+		// should be empty now
+		c := bucket.Cursor()
+		for k, v := c.First(); k != nil; k, v = c.Next() {
+			b.Errorf("unexpected key: %x = %q", k, v)
+		}
+		return rollback
+	}
+	if err := db.Update(validate); err != nil && err != rollback {
+		b.Error(err)
+	}
+}
+
+func BenchmarkDBBatchAutomatic(b *testing.B) {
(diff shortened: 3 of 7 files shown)


benbjohnson:
@tv42 I definitely like the idea of pushing it into Bolt core. It's such a common use case that it doesn't make sense to push it into another repo. We can't have it replace `DB.Update()` because the idempotency piece will cause backwards incompatibility.

I like your implementation better since it does the retries instead of rolling back everything. I don't have time right this moment to give it enough testing but I'll test and merge it in within the next couple of days.

tv42:
Yeah, `DB.Batch()` would be better than `DB.Update()`.

tv42:
Rebased and edited the commit to become DB.Batch.

davecgh:
I would highly recommend this pull request include documentation additions to README.md with example usage, caveats, etc.

tv42:
Needs more test coverage too. That work is just not useful if the whole approach is unwanted.

arctica:
Excellent addition. I hope this can be merged into Bolt core and the documentation expanded as davecgh suggested.

benbjohnson:
Sorry for the delay but I wanted to make sure I had a couple hours to dedicate to the PR and really understand it better. Here's some feedback after playing with it for a while:

1. What's the benefit of catching `panic()` calls and re-panicing? It seems like we'd lose the stacktrace of where the panic originated from.

2. The loop/`LoadPointer()`/`CompareAndSwapPointer()` in `DB.Batch()` is really confusing. I understand it conceptually but it feels clever and that worries me. Can we simplify this to kick off a new batch into a separate goroutine and add calls to that? Adding a call once full would cause it to flush and generate a new batch.

3. Can you rename `BatchMaxSize` and `BatchMaxDelay` to `MaxBatchSize` and `MaxBatchDelay`? That's just a personal preference.

4. Can you move `MaxBatchSize`/`MaxBatchDelay` to exported properties on `DB` instead of passing in through `Options`? The `Options` exists to set options that are required during `Open()`.

Also, I refactored some of the tests a bit:

https://github.com/benbjohnson/bolt/commit/ccee92e537d942fd763f2b2677bf52900c48fa39

tv42:
1. Good catch.

I can't just let the panic bubble up like Update or View do, because the func is typically executed in a different goroutine than where the matching Batch call is. The error has to be shipped to the right caller (hence the channel return from merge), and the type `panicked` is just used as a marker that it was a panic not an error.

Second, the point of re-panicking instead of just returning an error is to keep the panic, well, exceptional. It just felt like the right thing; don't swallow the panic because we actually have a caller that can deal with it (as opposed to e.g. a net/http Handler).

Maybe the recover bit should store the stack trace, and store it in the error sent to the caller. runtime.Stack makes it easy. (All goroutines or just current goroutine?) This seems to be the right thing to do regardless of whether it's re-panicked or returned as an error.

Let me know what you think.

2. It's essentially locking so it is bound to be clever by association. (Well, a full mutex would work too, but this oughta be faster). I wrote it originally with a goroutine but liked that even less; the atomic swap is the trick that makes it so nicely able to opportunistically join an existing batch, the nil->me (or prevFullBatchIJustObserved->me) transition with CompareAndSwapPointer can only happen one at a time so that's a good signal for "I should become master". And once I added that, the need for the goroutine and channel just naturally went away.

I'm very much open to making the code better, but I wonder if that's achievable with e.g. a nicer description of the state machine.

3. Sure. I was trying to group by being related to Batch.

4. I did it this way to avoid the situation where callers can change the values on the fly, both for races and the weird bugs that could trigger. I'm a little worried about leaving that open. This also let me do error handling up front. I understand that that is what Options currently does, but if you look at it with fresh eyes nothing about Options itself seems specific to the actual opening event; it's just configuration passed to a constructor function.

I can't vouch for the test changes right now, but I appreciate the motivation; less indentation is good, I was copying your older programming style ;)

tv42:
The test changes look good. Hint for reviewing: make diff ignore whitespace changes.

tv42:
batch-control-flow

```dot
digraph batch_control_flow {
  start [shape=point];
  testCur [shape=diamond, label="is there a batch?"];
  start -> testCur;
  
  tryJoin [shape=diamond, label="try to join\ncurrent batch"];
  testCur -> tryJoin [label="yes"];
  waitForErr [label="wait for error"];
  tryJoin -> waitForErr [label="joined,\nif full\nnotify master"];
  returnOrPanic [label="return or panic", shape=doublecircle];
  waitForErr -> returnOrPanic;
  tryJoin -> tryMaster [label="join refused\n(already started)"];
  
  tryMaster [shape=diamond, label="try to\nbecome\nmaster"];
  testCur -> tryMaster [label="no"];
  tryMaster:se -> testCur:n [label="lost race"];
  waitFullOrDelay [label="wait for full or max delay"];
  tryMaster:sw -> waitFullOrDelay [label="we are master"];
  doTransactions [label="call db.Update, handle errors"];
  waitFullOrDelay -> doTransactions [label="full or timer /\nmark as started"];
  doTransactions -> waitForErr;
}
```

alvarolm:
First of all thanks @benbjohnson @tv42 and everyone who contributed to the project for their invaluable efforts, initiatives like this make golang and their community shine, I love the simple api and guarantees like serializable transactions and ACID semantics, incredible useful for applications that require to handle data in a safe manner, thanks again !

imho I believe this feature is a huge step forward, in the mean time I'm still using the coalesce, so in the same line I wonder if the actual implementation, which returns a rollback error to the previous operations is really necessary, why don't just use "continue", and wait for the transaction result to return ?

tv42:
Alright for your reviewing pleasure, https://github.com/tv42/bolt/tree/batch-3 / 29c07645d27ab172e9c97bfee5443d5ecd7a3005 contains a rewrite using just a mutex for coordination. It got really simple once I realized I can wire a timer to a sync.Once.

For actual merging, I squashed the branch as 6931794f45b526f2cae8a84542d9544b049d9694.

It could still use more docs and tests, but is the current approach more understandable?

Performance is close enough to the atomics version, I'm seeing a lot of jitter so any real difference is probably drowned by the noise and would need several repeats to get anything conclusive; I'd rather get the code out there.

tv42:
Err I just saw a bug in handling batch full, consider the details to be in flux. Need those tests ;)

tv42:
The fix was pretty trivial: ee8e42e7c9bb44f278dfab6f5f16400f0e171445 (but once again, I don't think these intermediates are worth merging).

New squashed merge candidate is 4ce6f86e8846afbbe7d3c57390b237e9c844b766.

Coverage looks pretty good, the only thing missing is panicked.Error, and that's there only to make it fit through chan error.

tv42:
And I should still rip out the checks & errors on bad MaxBatchSize, MaxBatchDelay values and just handle whatever comes; if the user can set the fields on DB directly, there's no validation possible at that time anyway, and they degrade nicely in a way that just disables batching.

j:
:+1: and :smile:

tv42:
I'm rebasing the "batch" branch again, new commit is b906f4ecea67e2d52adc7ff3ffbbb8b4ddc05f83.
Here's what's new (from before the rebase, archived in branch batch-6):

commit 63c07803a9fe08343bd4cf84361f94f838569974

    Move batch item retrying to happen the original goroutine
    
    This preserves panic stack traces, as we no longer need to communicate
    panics as special errors between goroutines.

commit e6983837874b4da544e553ba5c569868ebf4b79c

    Internal docs for batch

commit d90b13ee45d025f2c63491dfe68226944ba06da1

    Enforce MaxBatchSize
    
    Previous logic relied on the scheduler to start the batch "soon
    enough", this one will never exceed MaxBatchSize.
    
    Fix broken test.

commit 0fabc638c631f5dd003f432489d1ae226dbf8109

    Saner batch tunables
    
    No need for errors here, and the copying logic was a leftover from
    when they were actually in Options.

tv42:
@benbjohnson what do you think of b906f4ecea67e2d52adc7ff3ffbbb8b4ddc05f83 / 5261724cfee3627e4198ed48bcaf18d949df3b3e?

@davecgh See the README addition in 5261724cfee3627e4198ed48bcaf18d949df3b3e, does this match what you wanted to see? If not, can you talk more about what you would need as a user.

README.md
```diff
@@ -125,6 +125,45 @@ no mutating operations are allowed within a read-only transaction. You can only
...
+
+For example: don't display messages from inside the function, instead
+set variables in the enclosing scope:
+
+```go
+var id uint64
+err := db.Batch(func(tx *bolt.Tx) error {
+	// Find last key in bucket, decode as bigendian uint64, increment
+    // by one, encode back to []byte, and add new key.
+    ...
```
davecgh: Inconsistent indentation here.

batch_benchmark_test.go
```diff
@@ -0,0 +1,126 @@
...
+	buf := make([]byte, 4)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		start := make(chan struct{})
+		var wg sync.WaitGroup
+
+		for round := 0; round < 1000; round++ {
+			wg.Add(1)
+
```
davecgh: Minor nitpick: This function has a blank line here while the other two that are virtual identical do not.

batch.go
```diff
@@ -0,0 +1,132 @@
...
+					failIdx = i
+					return err
+				}
+			}
+			return nil
+		})
+
+		if failIdx >= 0 {
+			// take the failing transaction out of the batch. it's safe to
+			// shorten b.calls here because b.started has been set.
```
davecgh: This comment appears to reference the old `b.started` variable which no longer applies.

tv42: Good catch, fix is 477bcbc145ce555b66699780d74473e98c1e1e7a

batch_benchmark_test.go
tv42: Unified in 94a4c2681b7e6df87f0e588017ca14afe45ae9e8

davecgh:
@tv42 That addition to the README looks pretty good.  I think it might benefit from a couple of extra things:

* An example that shows it being used concurrently.  It really doesn't do anything differently (from a caller perspective) than `Update` when called as shown in the `README` which I think could easily lead to confusion for a user unfamiliar with the package.  This is especially true when coming at it from the perspective of leveldb batches which are asynchronous calls that essentially just add the data to a batch of operations which you specifically trigger with a call to perform when ready (essentially what bolt already does inside of `Update`).
* A short discussion of when you wouldn't want to use a Batch.  For example, given the previous point, calling it from a single goroutine is semantically equivalent to calling `Update`.  Since a batch involves additional overhead, it might make sense to specifically call out that `Update` should be called instead unless you specifically arrange/need to call `Batch` in a concurrent fashion.

tv42:
@davecgh Thank you.

The problem with actually showing concurrency in the README is that it just gets too verbose. That would have to be a separate example.

Batch is definitely aimed at the case where the code is e.g. serving incoming network requests with multiple goroutines; anything else cannot fit into the same "returns error" calling pattern, and then you might as well use Update or manual transactions.

663b133e823e37b38b22eafd6b260790469f9d81 adds "concurrent" in key spots and explicitly says "Batch is only useful when there are multiple goroutines calling it" -- how's that?

davecgh:
@tv42 Looks good to me with those additions.

Agreed about the example directly in the README.  I was thinking more along the lines of referencing an example that is done in in `db_test.go`.  For example, `func ExampleDB_Batch()` similar to the existing [Update](https://github.com/boltdb/bolt/blob/master/db_test.go#L464-L493) example

tv42:
@davecgh I still have trouble coming up with a non-silly example that's brief enough. Usually concurrency is driven by communication with the outside world, but that's not something you can encapsulate in ExampleDB_Batch easily.

tv42:
@davecgh b8ad031bf17bf4d887ec0dcbda715d1ef9e27a45 has an example for Batch. I couldn't come up with anything simpler that looked like actual use.

j:
Can't wait.  Example looks clear enough.

benbjohnson:
@tv42 Sorry for the delay in re-reviewing this. I'm going to block out some time tonight to go through this PR again. Thanks again for all the work on it. It'll be nice to have batching in Bolt core.

tv42:
Rebased and squashed with all of the above nits handled (as far as I can tell) as  commit 1578db9cc2d71a8e7e6c16ff9a4401c89a0c3a7d. If there's any more feedback, please let me know.

batch.go
```diff
@@ -0,0 +1,135 @@
...
+					failIdx = i
+					return err
+				}
+			}
+			return nil
+		})
+
+		if failIdx >= 0 {
+			// take the failing transaction out of the batch. it's
+			// safe to shorten b.calls here because b.batch no longer
```
benbjohnson: `b.batch` -> `db.batch`

db.go
```diff
@@ -27,6 +27,14 @@ const magic uint32 = 0xED0CDAED
 // must be synchronzied using the msync(2) syscall.
 const IgnoreNoSync = runtime.GOOS == "openbsd"
 
+// Default values if not set in a DB instance.
+//
+// Do not change while databases are open.
+var (
+	DefaultMaxBatchSize  int = 1000
+	DefaultMaxBatchDelay     = 10 * time.Millisecond
+)
```
benbjohnson: What about making these constants since the `DB` fields are variable? If you change these to constants then you won't need the "do not change" warning above.

benbjohnson:
@tv42 I like the changes a lot. I think it flows really well now and it's easy to read. I like the use of `sync.Once` too. I made a few minor comments. The only other things I'd like to see are:

1. Validate the inserted data at the end of the benchmarks to make sure that they are correct. We can call `b.StopTimer()` to exclude that time to validate.

2. Add a `testing/quick` test to try a lot of different settings and validate that they all work as expected. I can try to add this in the next couple of days (or feel free to add it yourself). It'd be nice to run `go test -quickchecks=1000000` and just let it run for a while to make sure there's not any unexpected behavior.

tv42:
I just realized the benchmarks are racy, they share hashing across goroutines. Fixing..

tv42:
Alright fixes for the above are in commits leading to e927eff317cb2f8640cca95129a1faabb754b496 and I rebased & squashed on top of the new master as 15389f9c3de8b7c870279eb3e3599bcab416cdc9.

tv42:
Somewhere along the line I screwed up the commit message, fixed in adbb1a19c1b93a95f17a4eb3c5524e5be8b0d10f

tv42:
Ping.

benbjohnson:
@tv42 Sorry for the delay on this. I keep wanting to add some `testing/quick` tests but I haven't had the time. Just to confirm, is this PR ready for merge?

tv42:
@benbjohnson Apart from deserving more testing coverage -- which is just about always true -- I feel like I have nothing more to add, and the comments from others have stopped too.

PR #286, opened 2015-01-18
Add truncate() and sync() on resize.

### LINKED ISSUE #284: ext3/ext4 is not fully POSIX, to be safe there, need to fsync after file size changes
@cespare pointed out this conversation to me: http://www.openldap.org/lists/openldap-devel/201411/msg00000.html (I've seen the talk before, but didn't pay enough attention).

Reading http://linux.die.net/man/2/fdatasync says (my emphasis) "fdatasync() [...] does not flush modified metadata **unless that metadata is needed** in order to allow a subsequent data retrieval to be correctly handled. [...] On the other hand, a change to the file size (st_size, as made by say ftruncate(2)), would require a metadata flush."

Quoting the paper (my emphasis): "On ext3 with the default “ordered” journaling mode, the file data is forced directly out to the main file system **prior to its metadata** being committed to the journal. This is why we observe the journaling of the length update (op#399, 400, and 402) after the file data updates(op#342–398)." "[...] means fdatasync on ext3 does not wait for the completion of journaling (similar behavior has been observed on ext4)."

Currently, bolt seems to rely on individual page writes increasing the st_size of the file; there's no file size change where it actually notices it needs more space:
https://github.com/boltdb/bolt/blob/15a58b04ae5336fd58e2dab6f289459941fbaed6/db.go#L554

So, my takeaway from this is, ext3, and likely ext4, have a bug / "design tradeoff". If bolt wants to be pragmatic it should probably accommodate for that bug.

I think it would be enough to do something like this:

```go
	if minsz >= db.datasz {
		if err := db.file.Truncate(minsz); err != nil {
			return nil, fmt.Errorf("file resize error: %s", err)
		}
		if err := db.file.Sync(); err != nil {
			return nil, fmt.Errorf("file sync error: %s", err)
		}
		if err := db.mmap(minsz); err != nil {
			return nil, fmt.Errorf("mmap allocate error: %s", err)
		}
	}
```

Doing the resize once, and not page-by-page when writing out dirty pages, might even be more efficient.

benbjohnson:
## Overview

This commit fixes an issue with ext3/ext4 filesystems where metadata file size is not synced when resizing a file. It also resizes the entire resize instead of updating the size during individual page writes.

Fixes #284.

Thanks to @tv42 for the fix.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -3,6 +3,7 @@
 package bolt
 
 import (
+	"fmt"
 	"os"
 	"syscall"
 	"time"
@@ -41,6 +42,16 @@ func funlock(f *os.File) error {
 
 // mmap memory maps a DB's data file.
 func mmap(db *DB, sz int) error {
+	// Truncate and fsync to ensure file size metadata is flushed.
+	// https://github.com/boltdb/bolt/issues/284
+	if err := db.file.Truncate(int64(sz)); err != nil {
+		return fmt.Errorf("file resize error: %s", err)
+	}
+	if err := db.file.Sync(); err != nil {
+		return fmt.Errorf("file sync error: %s", err)
+	}
+
+	// Map the data file to memory.
 	b, err := syscall.Mmap(int(db.file.Fd()), 0, sz, syscall.PROT_READ, syscall.MAP_SHARED)
 	if err != nil {
 		return err


benbjohnson:
@tv42 how does this look?

tv42:
Can't see anything wrong with it. As the paper pointed out, there's a devil in every detail ;)

benbjohnson:
cool, thanks for looking it over.

PR #287, opened 2015-01-18
fix 'range scans' example

justphil:
Due to the fact that you want to iterate over all keys that are before or equal to `max` starting from `min` the bytes.Compare() check should look like the commit suggests.


benbjohnson:
@justphil cool, thanks for catching that. :+1:

PR #290, opened 2015-01-22
Removed redundant duplicate line of code

pmcgrath:
Presume it was a mistake, doesn't do any harm but is redundant code

### DIFF
--- a/db.go
+++ b/db.go
@@ -255,7 +255,6 @@ func (db *DB) init() error {
 		m.magic = magic
 		m.version = version
 		m.pageSize = uint32(db.pageSize)
-		m.version = version
 		m.freelist = 2
 		m.root = bucket{root: 3}
 		m.pgid = 4


benbjohnson:
The `version` field in `meta` stores the current database version to the underlying database. The Bolt data format isn't going to change anymore but it's useful for anyone who may have an old Bolt database when Bolt was still in alpha.

pmcgrath:
Maybe i'm missing something ? Hope i'm not wasting your time

I meant that line 256 and 258 in db.go are duplicates, line 258 is just re-setting the meta value with the same value again ? see https://github.com/boltdb/bolt/blob/master/db.go#L256-258 

 Thanks
Pat

benbjohnson:
Ah! Sorry about that. I misunderstood and read the code change too quickly. I didn't realize the lines were literally duplicated. I thought you meant setting the version wasn't necessary since it's already declared. lgtm. :+1: thanks!

Issue #291, opened 2015-01-24
DB size doubling on OS X


kemist:
I'm on OS X Yosemite 10.10.1, using tip. go version go1.4.1 darwin/amd64

When creating a DB larger than 4096 KiB, future calls to `bolt.Open/Close` will double the file size of the DB. Probably related to #216. Strangely the tests for 216 were dropped after a refactor, https://github.com/boltdb/bolt/commit/ba6badc#diff-65348100fd3ccd8cb8fb4f6bad6bc2d9L362

Many thanks for all your work on Bolt!

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/boltdb/bolt"
)

const path = "testdb.db"

func main() {
	// When Bolt is under 4096 (pagesize), everything is cool
	fmt.Println("Add one key, DB < 4096 bytes")
	seedTestDB(1)
	fmt.Printf("%d KiB\n", dbSize()) // 4096 KiB
	openAndClose()
	fmt.Printf("%d KiB\n", dbSize()) // 4096 KiB

	fmt.Println("")

	// If Bolt is > 4096 bytes, any subsequent open/close will double its size
	seedTestDB(1000)
	fmt.Println("Add 1000 keys, DB > 4096 bytes:")
	fmt.Printf("%d KiB\n", dbSize()) // 8208 KiB
	openAndClose()
	fmt.Printf("%d KiB\n", dbSize()) // 16416 KiB

}

func seedTestDB(numKeys int) {
	db, err := bolt.Open(path, 0600, nil)
	defer db.Close()
	if err != nil {
		log.Fatal(err)
	}

	err = db.Update(func(tx *bolt.Tx) error {
		pings, err := tx.CreateBucketIfNotExists([]byte("testbucket"))
		if err != nil {
			return err
		}
		for x := 0; x < numKeys; x++ {
			key := []byte(fmt.Sprintf("%d", x))
			val := make([]byte, 4096)

			err = pings.Put(key, val)
			if err != nil {
				return fmt.Errorf("Error writing key: %s", err)
			}
		}

		return nil
	})

	if err != nil {
		log.Fatal(err)
	}
}

func openAndClose() {
	db, err := bolt.Open(path, 0600, nil)
	defer db.Close()
	if err != nil {
		log.Fatal(err)
	}
}

func dbSize() int64 {
	file, err := os.Open(path)
	if err != nil {
		log.Panic(err)
	}
	defer file.Close()

	stat, err := file.Stat()
[...]

PR #292, opened 2015-01-28
Fix mmap resize calculation.

benbjohnson:
## Overview

This pull request fixes an issue where the database would grow whenever it was opened. This was caused by a recent change that performed a truncation when the database grew. Now there are fixed growth sizes for the database (1MB, 2MB, 4MB, 8MB, etc) up to 1GB and then the database will grow by 1GB when it resizes.

See also: 6bb25854a183f3d3bfa50096f910d3a3984e9834

Fixes: #291

/cc @kemist

### DIFF
--- a/db.go
+++ b/db.go
@@ -225,24 +225,30 @@ func (db *DB) mmapSize(size int) (int, error) {
 		}
 	}
 
-	// Verify the map size is not above the maximum allowed.
-	if size > maxMapSize-maxMmapStep {
+	// Verify the requested size is not above the maximum allowed.
+	if size > maxMapSize {
 		return 0, fmt.Errorf("mmap too large")
 	}
 
 	// If larger than 1GB then grow by 1GB at a time.
-	size += maxMmapStep
-	if remainder := size % maxMmapStep; remainder > 0 {
-		size -= remainder
+	sz := int64(size) + int64(maxMmapStep)
+	if remainder := sz % int64(maxMmapStep); remainder > 0 {
+		sz -= remainder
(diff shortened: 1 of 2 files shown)


benbjohnson:
@tv42 Does this look correct to you? It was caused by the truncation-on-resize change I made recently.

benbjohnson:
@tv42 Good call. Fixed: 4e3fb8d

tv42:
Oh good catch, now it actually changes the size on the disk and not just the mmap view, and the old logic had a self-feeding spiral..

I added comments to https://github.com/benbjohnson/bolt/commit/834b38e3e742f99b682311f61afc223407eca757, summary: off-by-one, 32-bit worries, me nagging about errors.

benbjohnson:
Awesome, thanks for looking through. I made the max step change and max size check change here: dacc187

benbjohnson:
OK, I spent 2 hours trying to get a 32-bit vagrant working but no ****ing luck. Here's my next attempt. I'm converting to int64 to prevent overflow and checking. Does this make sense: 338d8e7

tv42:
@benbjohnson Yeah can't see anything wrong with this one. Sometimes the simplest things are the hardest ones ;)

benbjohnson:
Awesome, thanks! By the way, good luck on the Bolt talk at GoSF. Is it being recorded?

tv42:
@benbjohnson All I can say at this time is that I've seen earlier ones recorded, and sometimes they show up on youtube.

PR #294, opened 2015-01-31
Expand assertion statements

benbjohnson:
## Overview

This pull request expands calls to `_assert()` that use variadic arguments. These calls require conversion to `interface{}` so there was a large number of calls to Go's internal `convT2E()` function. In some profiling this was taking over 20% of total runtime. I don't remember seeing this before Go 1.4 so perhaps something has changed. Or maybe I just didn't see it.

Note that this does not change any functionality. The changed lines of code are functionally equivalent.

---

/cc @mkobetic


mkobetic:
:+1:

PR #300, opened 2015-02-11
Avoid unnecessary locking

jedborovik:
Very minor change here. Just came across this and it seemed there's no need to lock if `db` isn't opened.


benbjohnson:
You still need to obtain the `metalock` to check against `db.opened`. Otherwise you'll get errors from the race detector.

PR #309, opened 2015-02-18
added gocode.io to list of projects using bolt

everdev:


benbjohnson:
Awesome! Thanks for PR'ing onto the list.

PR #319, opened 2015-02-28
Added max alloc size to arm.

skirmish:
Added max alloc size to the arm arch, because I was running bolt on a Pi2 and it wouldn't compile.


benbjohnson:
:+1: awesome, thanks!

PR #326, opened 2015-03-07
Add DeleteBucketIfExists

davidcornu:
Thought this might be useful. I've found that in most cases where I want to delete a bucket, I don't really care if it's actually there or not.

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -250,6 +250,15 @@ func (b *Bucket) DeleteBucket(key []byte) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


benbjohnson:
@davidcornu Thanks for submitting a PR. My biggest concern is that it clutters the API (which people already complain is too large) and the check for `ErrBucketNotFound` is pretty easy. The `CreateBucketIfNotExists()` is only there because initializing a set of top-level buckets on start up is so common in Bolt projects.

davidcornu:
:+1:

PR #330, opened 2015-03-17
Implement io.WriterTo interface on Tx.

benbjohnson:
## Overview

This pull request moves the functionality in `Tx.Copy()` to `Tx.WriteTo()`. This allows `Tx` to be used as an `io.WriterTo` which makes it easier to mock.

The `Tx.Copy()` function still exists but it's simply a wrapper around `Tx.WriteTo()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ func (*Bucket) DeleteBucket(key []byte) error
 
 ### Database backups
 
-Bolt is a single file so it's easy to backup. You can use the `Tx.Copy()`
+Bolt is a single file so it's easy to backup. You can use the `Tx.WriteTo()`
 function to write a consistent view of the database to a writer. If you call
 this from a read-only transaction, it will perform a hot backup and not block
 your other database reads and writes. It will also use `O_DIRECT` when available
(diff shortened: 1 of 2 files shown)


baijum:
:thumbsup:

How about updating `Database backups` section in the `README.md` to use this new API?

benbjohnson:
Good catch! I updated the docs in 8c6af54.

README.md
```diff
@@ -343,7 +343,8 @@ func BackupHandleFunc(w http.ResponseWriter, req *http.Request) {
 		w.Header().Set("Content-Type", "application/octet-stream")
 		w.Header().Set("Content-Disposition", `attachment; filename="my.db"`)
 		w.Header().Set("Content-Length", strconv.Itoa(int(tx.Size())))
-		return tx.Copy(w)
+		_, err := tx.WriteTo(w)
+		return err
```
rhcarvalho: What's the benefit of writing 2 lines to use `tx.WriteTo` instead of the original 1 line with `tx.Copy`?

benbjohnson: There's no benefit, necessarily. I mainly like to stick with the stdlib interfaces where possible.

rhcarvalho: :+1: thanks for answering.

PR #333, opened 2015-03-24
Document key/value safety.

benbjohnson:
## Overview

This commit adds safety documentation to the data accessor functions in Bolt as well as the README. This was documented once in the package level godoc but it's important enough that it should be more clear.

---

/cc @mschoch


benbjohnson:
:dancer:

Issue #350, opened 2015-04-14
Allowing db files under 1MB


mcuadros:
I am using bolt as storage engine for [raa](https://github.com/mcuadros/go-raa), since is a general purpose archive format, in some cases the contained files could be less than 1MB.

I wish know if this bootstrapping size can be configured on some way per database, allowing db files under 1MB. This change could have performance impact, due to the continues growing mmap?

Thanks!

chrsm:
@benbjohnson Looks like this one can be closed with #351 fixed.

benbjohnson:
:+1:

PR #351, opened 2015-04-14
db: MinMmapSize is the minimum mmap size in bytes expressed in base 2

mcuadros:
As https://github.com/boltdb/bolt/issues/350 says in some cases the optimization to force the minimum file size to 1M is not very useful when your database is very small.
 
This PR introduce the Option.MinMmapSize param to allow lower sizes


benbjohnson:
Sorry for the delay. The `1MB` minimum was pretty arbitrary. I'd actually be ok with just setting the min to `32KB`. The minimum size is 20KB (2 meta pages, 1 free list page, 1 root page, and 1 bucket root page) so `32KB` is the next highest power of `2`.

Does that work for you?

mcuadros:
Yes, actually I fixed the value to 2, and in benchmarks work pretty similar. Thanks

benbjohnson:
@mcuadros I made the change to the minimum mmap size here: https://github.com/boltdb/bolt/pull/375

PR #363, opened 2015-05-05
Add DB.NoTruncate flag.

benbjohnson:
## Overview

This commit adds the `DB.NoTruncate` flag to optionally revert `mmap()` calls to how they were implemented before the ext3/ext4 fix. When `NoTruncate` is true, remapping the data file will not force the file system to resize it immediately. This works for non-ext3/4 file systems.

The default value of `NoTruncate` is `false` so it is still safe for ext3/ext4 file systems by default.

See also: https://github.com/boltdb/bolt/issues/284

/cc @tv42 @mkobetic @bouk

### DIFF
--- a/db.go
+++ b/db.go
@@ -55,6 +55,14 @@ type DB struct {
 	// THIS IS UNSAFE. PLEASE USE WITH CAUTION.
 	NoSync bool
 
+	// When true, skips the truncate call when growing the database.
+	// Setting this to true is only safe on non-ext3/ext4 systems.
+	// Skipping truncation avoids preallocation of hard drive space and
+	// bypasses a truncate() and fsync() syscall on remapping.
+	//
+	// https://github.com/boltdb/bolt/issues/284
+	NoGrowSync bool
+
 	// MaxBatchSize is the maximum size of a batch. Default value is
 	// copied from DefaultMaxBatchSize in Open.
 	//
@@ -123,6 +131,7 @@ func Open(path string, mode os.FileMode, options *Options) (*DB, error) {
 	if options == nil {
 		options = DefaultOptions
 	}
+	db.NoGrowSync = options.NoGrowSync
 
 	// Set default values for later DB operations.
 	db.MaxBatchSize = DefaultMaxBatchSize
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -55,6 +55,12 @@ type DB struct {
 	// THIS IS UNSAFE. PLEASE USE WITH CAUTION.
 	NoSync bool
 
+	// When true, skips the truncate call when resizing the database.
+	// Setting this to true is only safe on non-ext3/ext4 systems.
+	//
```
mkobetic: Would it be worth adding a note here the we found that Truncate can be very slow on some filesystems?

benbjohnson: I wasn't able to determine the root cause. Truncation did work very quickly in some cases but in others it was taking >10m. It may have been a configuration issue or a hardware issue so I'd rather not assign blame to the FS.

Also, `NoTruncate` is nice if you don't want to preallocate space so I'll list that as a benefit.

benbjohnson: Updated in 7a65f95.

tv42:
Can we call it something else than NoTruncate? First of all, the word truncate implies shortening, though it's the syscall used to adjust file size either way; here, it's used for growing. And the cost really comes from the sync.

Since technically it is, or might be depending on how to interpret the POSIX bible, ext3/ext4 etc fault, I am not going to recommend `UnsafeMetadata`; as far as I understand, it's *supposed* to be safe without the truncate+sync, but it's not. `NoMetadataSync`?

benbjohnson:
I agree that `NoTruncate` sounds weird. The `truncate()` call is the one that is inexplicably hanging for us.

I have two issues with naming it `NoMetadataSync`:

1. LMDB implements `NOMETASYNC` which sounds the same but works differently.

2. It's only related to remapping. The name makes it sound like it always skips metadata sync.

How about `NoRemapSync`?

tv42:
`NoGrowSync`? `NoSyncAfterGrow`? `NoSizeSync`?

benbjohnson:
I think `NoGrowSync` would be my favorite from those three. Why not `NoRemapSync`?

tv42:
Because it's not really about mmap, either. Off hand, I think it even triggers on the first mmap, so it's really not a remap. The purpose of that code is to sync the new, larger, size of the file to disk.

benbjohnson:
Changed to `NoGrowSync`. Merging. Thanks @tv42!

Issue #364, opened 2015-05-07
Add db.Sync() api


xiaost:
hello, @benbjohnson 
In my project, I need something like "sync every second"
boltdb only provides `db.NoSync` option, 
Eh...I can do it by committing an empty tx with `db.NoSync = true` every second :cry: 
It looks good to add `db.Sync()` api, right?

benbjohnson:
Added `DB.Sync()` to https://github.com/boltdb/bolt/pull/366. Can you take a look and verify that's what you're looking for?

xiaost:
Yeah, exactly what I need. :+1: :laughing:

benbjohnson:
Cool, it's merged. :+1:

PR #366, opened 2015-05-08
Add DB.Sync()

benbjohnson:
## Overview

This pull request adds `DB.Sync()` for users who set `DB.NoSync` to `true`. This ensures `NoSync` commits are safe after syncing but it still provides no guarantees to `NoSync` before syncing.

Note: Please use `NoSync` carefully. Partially written commits can corrupt your database when using `NoSync`.

Fixes https://github.com/boltdb/bolt/issues/364

### DIFF
--- a/db.go
+++ b/db.go
@@ -556,6 +556,12 @@ func (db *DB) View(fn func(*Tx) error) error {
 	return nil
 }
 
+// Sync executes fdatasync() against the database file handle.
+//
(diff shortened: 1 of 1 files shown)


calmh:
> Note: Please use NoSync carefully. Partially written commits can corrupt your database when using NoSync.

Could you expand something on this (I.e. Linder what circumstances it can happen), and why this is not a problem when NoSync is false?

PR #371, opened 2015-05-14
open read-only databases in read-only mode

sasha-s:
allows a database to be opened by multiple processes at the same time in read-only mode.
all attempts to mutate the database result in ErrDatabaseReadOnly.

### DIFF
--- a/db.go
+++ b/db.go
@@ -140,14 +140,8 @@ func Open(path string, mode os.FileMode, options *Options) (*DB, error) {
 	db.MaxBatchSize = DefaultMaxBatchSize
 	db.MaxBatchDelay = DefaultMaxBatchDelay
 
-	// Get file stats.
-	s, err := os.Stat(path)
 	flag := os.O_RDWR
-	if err != nil && !os.IsNotExist(err) {
-		return nil, err
-	} else if err == nil && (s.Mode().Perm()&0222) == 0 {
-		// remove www from mode as well.
-		mode ^= (mode & 0222)
+	if options.ReadOnly {
 		flag = os.O_RDONLY
 		db.readOnly = true
 		// Ignore truncations.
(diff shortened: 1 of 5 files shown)


tv42:
Read-only use should take a read lock, to prevent concurrent writers.

sasha-s:
it could, but the concurrent writes are impossible because the file is read-only (one would not be able to open it for writing without doing chmod).

tv42:
@sasha-s Concurrent as in a separate process.

sasha-s:
@tv42 all writes are impossible if the file is read-only (concurrent or not), thus read lock is not needed.

tv42:
@sasha-s: But without the lock, nothing guarantees the file content stays stable.

sasha-s:
@tv42 how can the file content get changed?
some other process has to open it in write mode for that to happen. and it is impossible because the file is readonly.
BTW if by read lock you mean flock (could be shared flock in this case) - please note that flock is an advisory lock. From flock man page:
```
flock() places advisory locks only; given suitable permissions on a file, a process is free to ignore the use of flock() and perform I/O on the file.
```
relying on a file being read-only seems cleaner and more robust.

benbjohnson:
@sasha-s `ReadOnly` seems like a decision that should be made by the application explicitly and not by the file mode implicitly. For example, if it's determined by the file mode then an application wouldn't error out until it's first `Update()` which could take a while depending on the application. If it's an explicit decision by the application then it should error out on open.

I agree with @tv42 about `flock()`. I can see the situation where a database is infrequently updated but read in by multiple processes. Each process would have to `bolt.Open()` every time it reads data but a read lock would allow multiple processes to do that concurrently. @kelseyhightower wrote a bolt-backed app like that recently. There's no way to do that safely by turning permissions on and off.

sasha-s:
Please have another look @benbjohnson

tv42:
@sasha-s File access mode just isn't a good mechanism for knowing when the file is guaranteed to not change.

```console
[0 tv@brute ~]$ mkdir demo
[0 tv@brute ~]$ cd demo/
[0 tv@brute ~/demo]$ mkdir mnt
[0 tv@brute ~/demo]$ bolt-mount db mnt &
[1] 31933
[0 tv@brute ~/demo]$ mkdir mnt/bukkit
[0 tv@brute ~/demo]$ echo value >mnt/bukkit/key
[0 tv@brute ~/demo]$ ls -l
total 24
-rw------- 1 tv   tv   1048576 May 18 11:29 db
drwxr-xr-x 1 root root       0 May 18 11:29 mnt/
[0 tv@brute ~/demo]$ chmod a-w db
[0 tv@brute ~/demo]$ ls -l
total 24
-r-------- 1 tv   tv   1048576 May 18 11:29 db
drwxr-xr-x 1 root root       0 May 18 11:29 mnt/
[0 tv@brute ~/demo]$ echo more >mnt/bukkit/another
[0 tv@brute ~/demo]$ head mnt/bukkit/*
==> mnt/bukkit/another <==
more

==> mnt/bukkit/key <==
value
[0 tv@brute ~/demo]$ fusermount -u  mnt
```

tv42:
Ooh I see the stat aspect is gone. That's better. (This does not constitute a code review ;)

db.go
```diff
@@ -102,8 +102,11 @@ type DB struct {
 	statlock sync.RWMutex // Protects stats access.
 
 	ops struct {
-		writeAt func(b []byte, off int64) (n int, err error)
+		writeAt  func(b []byte, off int64) (n int, err error)
+		Truncate func(size int64) error
```
benbjohnson: This `ops` structure was added for mocking tests but didn't really get fleshed out enough. I'd rather see the check for `readOnly` done in each of the `mmap()` functions and then avoid `Truncate()` there if `true`. It's easier to read and there's not so much magic going on.

benbjohnson:
@sasha-s Overall the PR looks good except for one `ops` comment I made. Can you remove the `ops` function and make ignoring `Truncate()` explicit and then I'll do a full review and merge.

sasha-s:
Thanks @benbjohnson.
Please have yet another look.

benbjohnson:
Merged! Thanks @sasha-s. I added a section to the `README` as well: https://github.com/boltdb/bolt/commit/957d850b5158a4eebf915476058e720f43459584

PR #376, opened 2015-05-20
Add Consul to README

ryanuber:
[Consul](https://github.com/hashicorp/consul) has recently switched to using BoltDB as its backend store. It is used to store Raft logs to persist cluster state, and so far we love it! It has enabled us to shed the pain of compiling for multiple platforms with CGO. Keep up the great work, and thank you!


benbjohnson:
Woohoo! That's what I like to hear. :) I'm glad is working for you guys.

Issue #341, opened 2015-03-29
Wait to close DB if there are open transactions


timshannon:
This is probably an enhancment request, as I don't believe this behavior was unintentional, but you can open a DB, open a transaction, then close the DB.

```
package main

import (
	"fmt"
	"os"

	"github.com/boltdb/bolt"
)

func main() {
	// Open the database.
	db, _ := bolt.Open("test.db", 0666, nil)
	defer os.Remove(db.Path())

	// Start a write transaction.
	db.Update(func(tx *bolt.Tx) error {
		// Create a bucket.
		tx.CreateBucket([]byte("widgets"))
		b := tx.Bucket([]byte("widgets"))

		// Set the value "bar" for the key "foo".
		b.Put([]byte("foo"), []byte("bar"))

		// Retrieve the key back from the database and verify it.
		value := b.Get([]byte("foo"))
		fmt.Printf("The value of 'foo' was: %s\n", value)
		return nil
	})

	// Retrieve the key again.
	tx, _ := db.Begin(false)

	db.Close()
	value := tx.Bucket([]byte("widgets")).Get([]byte("foo"))
	if value == nil {
		fmt.Printf("The value of 'foo' is now: nil\n")
	}
}
```

The documention says that the value from the bucket is valid while the transaction is open, and in my mind, I see db.Close() as a kind of *safe shutdown*. 

There is a timeout option already for the db Open call, why not use the same timeout on Close, and wait for any currently open transactions to finish?

The copy on write nature of Bolt means you'll never have partially committed data, so the issue isn't one of integrity, but your client can end up with a panic.

benbjohnson:
I thought there was an open issue for this already but I'm looking back through the issues and I don't see one. This is definitely annoying since it can cause a panic. You're correct that it won't cause corruption though. I'll try to throw a PR together soon.

alaska:
Chiming in to suggest that Close() should wait on outstanding Batch() calls, too.  Seems like a ```waitgroup``` would be a perfect way to handle this.

benbjohnson:
Sorry for the delay. Added support for a blocking close here: https://github.com/boltdb/bolt/pull/377

benbjohnson:
I didn't add `Batch()` blocking because I wanted to make sure I got the basic use case working first.

PR #377, opened 2015-05-20
Wait for pending transactions on close

benbjohnson:
## Overview

This pull request fixes the `DB.Close()` function so that it waits for any open transactions to finish before closing.

Previously a close would shutdown the `DB` immediately and cause a panic in any open transactions, however, no data corruption would occur.

Fixes: https://github.com/boltdb/bolt/issues/341

### DIFF
--- a/db.go
+++ b/db.go
@@ -351,8 +351,15 @@ func (db *DB) init() error {
 // Close releases all database resources.
 // All transactions must be closed before closing the database.
 func (db *DB) Close() error {
+	db.rwlock.Lock()
+	defer db.rwlock.Unlock()
+
 	db.metalock.Lock()
 	defer db.metalock.Unlock()
+
+	db.mmaplock.RLock()
+	defer db.mmaplock.RUnlock()
+
 	return db.close()
 }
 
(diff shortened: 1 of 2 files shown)


db.go
```diff
@@ -351,8 +351,15 @@ func (db *DB) init() error {
...
 func (db *DB) Close() error {
+	db.rwlock.Lock()
+	defer db.rwlock.Unlock()
+
 	db.metalock.Lock()
 	defer db.metalock.Unlock()
+
+	db.mmaplock.RLock()
+	defer db.mmaplock.RUnlock()
+
```
benbjohnson: This locking order is consistent with the rest of the library (rwlock, metalock, mmaplock). The `mmaplock` takes a read lock because the `rwlock` prevents the write lock on `mmaplock`.

db_test.go
```diff
@@ -748,7 +791,7 @@ func (db *TestDB) PrintStats() {
 
 // MustCheck runs a consistency check on the database and panics if any errors are found.
 func (db *TestDB) MustCheck() {
-	db.View(func(tx *bolt.Tx) error {
+	db.Update(func(tx *bolt.Tx) error {
```
benbjohnson: This was changed because calling `Tx.Check()` on a read-only transaction is not thread safe.

Issue #378, opened 2015-05-21
deadlock with long-running read-transaction and write transaction on same go-routine


mschoch:
I first observed this in a bleve unti test failure.  It seems to occur when you have a long-running read-only transaction and a write transaction on the same go-routine.  I don't bleve actually needs this to work on a single go-routine, but it was written that way in a test for reader isolation.  The noteworthy part is that used to work (up until a450f843ee57645db062a4fe472c4509a7a1c7a2 if I used git bisect correctly).  Its possible this was never supposed to work, but might be interesting as it is a subtle change in behavior.  (its also not obvious to mewhy this particular commit resulted in this change)

I've written up a test-case outside of bleve that seems to replicate the same behavior:

```
package bolt

import (
	"io/ioutil"
	"os"
	"reflect"
	"testing"
)

func TestNewDeadlock(t *testing.T) {
	bucket := []byte("default")
	tmpdir, err := ioutil.TempDir("", "bolt")
	if err != nil {
		t.Fatal(err)
	}
	tmpfile := tmpdir + "/bolt"
	defer func() {
		os.RemoveAll(tmpfile)
	}()

	// open boltdb
	b, err := Open(tmpfile, 0600, nil)
	if err != nil {
		t.Fatal(err)
	}

	// create a default bucket
	err = b.Update(func(tx *Tx) error {
		_, err := tx.CreateBucketIfNotExists(bucket)
		return err
	})
	if err != nil {
		t.Fatal(err)
	}

	// write a kv pair
	err = b.Update(func(tx *Tx) error {
		return tx.Bucket(bucket).Put([]byte("a"), []byte("val-a"))
	})
	if err != nil {
		t.Fatal(err)
	}

	// create an isolated read tx
	r1, err := b.Begin(false)
	if err != nil {
		t.Fatal(err)
	}

	// verify we see the kv pair
	v1 := r1.Bucket(bucket).Get([]byte("a"))
	if !reflect.DeepEqual(v1, []byte("val-a")) {
		t.Fatalf("expected 'val-a', got '%s'", v1)
	}

	// write another kv pair
	err = b.Update(func(tx *Tx) error {
		return tx.Bucket(bucket).Put([]byte("b"), []byte("val-b"))
	})
	if err != nil {
		t.Fatal(err)
	}

	// make sure a new reader sees this kv pair
	b.View(func(tx *Tx) error {
		v2 := tx.Bucket(bucket).Get([]byte("b"))
[...]

PR #379, opened 2015-05-21
Add docs for dependent transactions.

benbjohnson:
## Overview

This pull request adds documentation to clarify that read-only and read-write transactions should not be mixed in the same goroutine as it can cause deadlocks during remapping.

See: https://github.com/boltdb/bolt/issues/378

/cc @mschoch


mschoch:
+1 looks good

PR #383, opened 2015-06-02
Add madvise() after mmap().

benbjohnson:
## Overview

This pull request advises the memory-mapped data file to use `MADV_RANDOM` to avoid readahead. This can provide a performance boost to Bolt databases that are larger than memory by avoiding unnecessary disk I/O.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -63,6 +63,11 @@ func mmap(db *DB, sz int) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


xiang90:
> performance boost 

Have you got a benchmark?

benbjohnson:
I don't have specific benchmark details that I can show but in production our disk latency went down by 80% and the number of pages being paged in & out dropped by about 90%.

PR #389, opened 2015-06-11
Fix meta copy invalid argument err

xiaost:
hi @benbjohnson . in my prj, `tx.WriteTo` returns "meta copy: invalid argument" err.

we use `bytes.Buffer` as argument of `tx.WriteTo`, 
since we may write some headers before calling `tx.WriteTo` the buffer not always page-aligned.
Under linux, `O_DIRECT` requires page-aligned buffer,
and `tx.WriteTo` should guarantee that.

test code:

``` go
buf := bytes.NewBuffer(make([]byte, 8))
ok(t, db.View(func(tx *bolt.Tx) error {
       n, err := tx.WriteTo(buf)
       return err
}))
```

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -274,7 +274,7 @@ func (tx *Tx) WriteTo(w io.Writer) (n int64, err error) {
 
 	// Copy the meta pages.
 	tx.db.metalock.Lock()
-	n, err = io.CopyN(w, f, int64(tx.db.pageSize*2))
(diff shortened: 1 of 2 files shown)


xiaost:
ping @benbjohnson , any other solutions?

benbjohnson:
@xiaost Sorry for the delay. I went through the code when you first PR'd but I want to research and understand the `O_DIRECT` issue better. I also want to do some additional `testing/quick` tests and try this on a variety of operating systems before merging. I just haven't had the time though to devote to it unfortunately.

PR #405, opened 2015-08-11
Fix invalid fmt.Fprintln/Errorf calls

artyom:
`go vet` discovered two calls of Fprintln with trailing newline and one
Errorf call without proper format string.


benbjohnson:
:+1: thanks!

PR #407, opened 2015-08-13
no need to call Truncate and Sync if size is not changed

funny-falcon:
https://github.com/boltdb/bolt/issues/284 mentions that Sync should be called on size change.
But if size is not changed, why Sync is called?
It is an issue for opening file: size is really the same.
Truncate+Sync slows Open very much: if 10 concurrent small files are opening, then speed drops from 20000open/sec to 50open/sec

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -45,10 +45,10 @@ func funlock(f *os.File) error {
 }
 
 // mmap memory maps a DB's data file.
-func mmap(db *DB, sz int) error {
+func mmap(db *DB, sz int, callTruncate bool) error {
 	// Truncate and fsync to ensure file size metadata is flushed.
 	// https://github.com/boltdb/bolt/issues/284
-	if !db.NoGrowSync && !db.readOnly {
+	if callTruncate {
 		if err := db.file.Truncate(int64(sz)); err != nil {
 			return fmt.Errorf("file resize error: %s", err)
 		}
--- a/db.go
+++ b/db.go
@@ -238,7 +238,8 @@ func (db *DB) mmap(minsz int) error {
 	}
 
 	// Memory-map the data file as a byte slice.
-	if err := mmap(db, size); err != nil {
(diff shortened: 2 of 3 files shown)


bolt_unix.go
```diff
@@ -45,10 +45,10 @@ func funlock(f *os.File) error {
 }
 
 // mmap memory maps a DB's data file.
-func mmap(db *DB, sz int) error {
+func mmap(db *DB, sz int, changesize bool) error {
```
benbjohnson: Can you change the variable name to `syncRequired`? I think it makes more sense within the context of the function.

db.go
```diff
@@ -238,7 +238,7 @@ func (db *DB) mmap(minsz int) error {
 	}
 
 	// Memory-map the data file as a byte slice.
-	if err := mmap(db, size); err != nil {
+	if err := mmap(db, size, int(info.Size()) != size); err != nil {
```
benbjohnson: Can you type convert `size` up to an `int64` instead of `info.Size()` down to an `int`?


e.g.

```
info.Size() != int64(size)
```

Theoretically a 32-bit machine could have a 4GB upper limit whereas `int` on a 32-bit machine has an upper limit of ~2B.

funny-falcon: ok, will do

bolt_unix.go
funny-falcon: ok

funny-falcon: Oh, I just found that I forgot about `bolt_windows.go` where is no call to `Sync`.
`Sync` looks to be helper method, and call to `Truncate` is the main.
So, I think `changeSize` is semantically better name.

More: calling to `Truncate` looks to also take non-zero time.

funny-falcon: well, I remade it completely: move decision logic to db.mmap

funny-falcon:
looks like it was fixed by https://github.com/boltdb/bolt/commit/e67705ed6348675b7b

PR #409, opened 2015-08-15
use sync.Pool for allocating small pages

funny-falcon:


benbjohnson:
Do you have benchmarks or memory profiles for this change?

funny-falcon:
I've added "-path" flag in db_test.go to specify where temporary db should be allocated.
I tested in '/dev/shm' (ie in-memory) cause I don't have SSD disk.
I get following times of `GOMAXPROCS=1 go test -path=/dev/shm`:
````
Go 1.4
before
ok      github.com/boltdb/bolt  195.516s
after
ok      github.com/boltdb/bolt  142.697s

Go 1.5
before
ok      github.com/boltdb/bolt  87.658s
after
ok      github.com/boltdb/bolt  79.913s
````
Go 1.5 really has much better GC collector. But sync.Pool helps here too.

Note, that sync.Pool plays well with GC: it is cleared before GC cycle. So that it can only decrease memory consumption, not increase.

funny-falcon:
It will also help when there is many open small bolt databases: then their page buffers will be reused between them.

LK4D4:
@benbjohnson I encountered this recently in Docker project. Using sync.Pool could help performance because of fewer allocations and also prevent too often GC collection. I don't see a drawback in this. Should we try again?
@funny-falcon are you still interested in this change?

benbjohnson:
I am interested in integrating sync.Pool but I want to get longer running integration tests and multi-platform CI running first.

funny-falcon:
Close in favor of #501

PR #410, opened 2015-08-16
Speedup open huge file

funny-falcon:
- use real set for freelist.cache
  map([pgids]struct{}) - is a real set structure, cause struct{} doesn't consume memory
- sort freelist.ids only if they aren't sorted
- do not read freelist if database opened in readonly mode

### DIFF
--- a/db.go
+++ b/db.go
@@ -115,10 +115,11 @@ type DB struct {
 	batchMu sync.Mutex
 	batch   *batch
 
-	rwlock   sync.Mutex   // Allows only one writer at a time.
-	metalock sync.Mutex   // Protects meta page access.
-	mmaplock sync.RWMutex // Protects mmap access during remapping.
-	statlock sync.RWMutex // Protects stats access.
+	rwlock       sync.Mutex   // Allows only one writer at a time.
+	metalock     sync.Mutex   // Protects meta page access.
+	mmaplock     sync.RWMutex // Protects mmap access during remapping.
+	statlock     sync.RWMutex // Protects stats access.
+	freelistonce sync.Once    // Protects reading freelist from file.
 
 	ops struct {
 		writeAt func(b []byte, off int64) (n int, err error)
@@ -230,12 +231,6 @@ func Open(path string, mode os.FileMode, options *Options) (*DB, error) {
 	if err := db.mmap(options.InitialMmapSize); err != nil {
 		_ = db.close()
 		return nil, err
-	}
-
-	if !db.readOnly {
-		// Read in the freelist.
-		db.freelist = newFreelist()
-		db.freelist.read(db.page(db.meta().freelist))
 	}
 
 	// Mark the database as opened and return.
@@ -441,6 +436,15 @@ func (db *DB) close() error {
 	return nil
 }
 
+func (db *DB) ensureFreelist() *freelist {
+	db.freelistonce.Do(func() {
+		fl := newFreelist()
+		fl.read(db.page(db.meta().freelist))
+		db.freelist = fl
+	})
+	return db.freelist
+}
+
 // Begin starts a new transaction.
 // Multiple read-only transactions can be used concurrently but only one
 // write transaction can be used at a time. Starting multiple write transactions
@@ -516,11 +520,11 @@ func (db *DB) beginRWTx() (*Tx, error) {
 	// Once we have the writer lock then we can lock the meta pages so that
 	// we can set up the transaction.
 	db.metalock.Lock()
-	defer db.metalock.Unlock()
 
 	// Exit if the database is not open yet.
 	if !db.opened {
 		db.rwlock.Unlock()
+		db.metalock.Unlock()
 		return nil, ErrDatabaseNotOpen
 	}
 
@@ -536,10 +540,14 @@ func (db *DB) beginRWTx() (*Tx, error) {
 			minid = t.meta.txid
 		}
 	}
+
+	// Release meta here cause ensureFreelist can take a long time
+	db.metalock.Unlock()
+
+	db.ensureFreelist()
 	if minid > 0 {
 		db.freelist.release(minid - 1)
 	}
-
 	return t, nil
 }
 
--- a/db.go
+++ b/db.go
@@ -232,9 +232,11 @@ func Open(path string, mode os.FileMode, options *Options) (*DB, error) {
 		return nil, err
 	}
 
-	// Read in the freelist.
-	db.freelist = newFreelist()
-	db.freelist.read(db.page(db.meta().freelist))
+	if !db.readOnly {
+		// Read in the freelist.
+		db.freelist = newFreelist()
+		db.freelist.read(db.page(db.meta().freelist))
+	}
 
 	// Mark the database as opened and return.
 	return db, nil
--- a/freelist.go
+++ b/freelist.go
@@ -175,7 +175,9 @@ func (f *freelist) read(p *page) {
 		copy(f.ids, ids)
 
 		// Make sure they're sorted.
-		sort.Sort(pgids(f.ids))
+		if !pgids(f.ids).isSorted() {
+			sort.Sort(pgids(f.ids))
+		}
 	}
 
 	// Rebuild the page cache.
--- a/freelist.go
+++ b/freelist.go
@@ -9,16 +9,16 @@ import (
 // freelist represents a list of all pages that are available for allocation.
 // It also tracks pages that have been freed but are still in use by open transactions.
 type freelist struct {
-	ids     []pgid          // all free and available free page ids.
-	pending map[txid][]pgid // mapping of soon-to-be free page ids by tx.
-	cache   map[pgid]bool   // fast lookup of all free and pending page ids.
+	ids     []pgid            // all free and available free page ids.
+	pending map[txid][]pgid   // mapping of soon-to-be free page ids by tx.
+	cache   map[pgid]struct{} // fast lookup of all free and pending page ids.
 }
 
 // newFreelist returns an empty, initialized freelist.
 func newFreelist() *freelist {
 	return &freelist{
 		pending: make(map[txid][]pgid),
-		cache:   make(map[pgid]bool),
+		cache:   make(map[pgid]struct{}),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


freelist.go
```diff
@@ -171,7 +174,11 @@ func (f *freelist) read(p *page) {
 	copy(f.ids, ids)
 
 	// Make sure they're sorted.
-	sort.Sort(pgids(f.ids))
+	for i, v := range f.ids[:len(f.ids)-1] {
+		if f.ids[i+1] < v {
+			sort.Sort(pgids(f.ids))
+		}
```
aybabtme: I think you want to `break` here, else you'll sort, then keep testing all following indices.

freelist.go
```diff
@@ -171,7 +174,11 @@ func (f *freelist) read(p *page) {
 	copy(f.ids, ids)
 
 	// Make sure they're sorted.
-	sort.Sort(pgids(f.ids))
+	for i, v := range f.ids[:len(f.ids)-1] {
```
aybabtme: seems equivalent to using https://golang.org/pkg/sort/#IsSorted:

```go
if !sort.IsSorted(pgids(f.ids)) {
   sort.Sort(pgids(f.ids))
}
```

freelist.go
funny-falcon: You're absolutely right

freelist.go
funny-falcon: Yes,  but IsSorted will be much slower. I'd rather use explicit loop here.

neurodrone: Won't [`sort.IsSorted`](https://golang.org/pkg/sort/#IsSorted) perform same amount of iterations as this? I think the advantage there is it will allocate no extra copies for each `pgid` that's iterated upon, unlike this implementation.

funny-falcon: @neurodrone pgid is just an integer, so that it's copy is very very cheap. And this custom loop will use direct integer comparisons.

sort.IsSorted will be much slower, cause it will call .Less on every iteration through interface description, and function call is many times slower than not doing function call.

neurodrone: That is true, although it does matter how many of those integers are copied around. Those copies are not necessary as far as I can see and can be avoided. It's similar to how in this PR we are proposing that bool values are more expensive compared to empty struct{}s, when in fact allocation for individual bools on a small scale is supposedly very cheap (on the order of a byte).

I think the assumption of function calls being expensive in Go doesn't necessarily stand true. The compiler inlines a ton of functions behind the scenes. Unless we can write a benchmark that proves calling a function (that only does a integer comparison) is indeed that expensive I think we should default to using the ones provided by stdlibs.

If we are worried about performance I think we should lean towards `sort.IsSorted` or at least prove this implementation is equivalent. However, If we don't care about it too much we can proceed with this implementation.

funny-falcon: > It's similar to how in this PR we are proposing that bool values are more expensive compared to empty struct{}s, when in fact allocation for individual bools on a small scale is supposedly very cheap (on the order of a byte).

But this PR is about huge scale: opening huge file. In huge scale there is the difference.

> If we are worried about performance (or readability) I think we should lean towards sort.IsSorted or at least prove this implementation is equivalent.

I agree about readability. But performance-wise `IsSorted` will be slower for at least couple of years, cause default Go compiller currently does no inlining of interface call,  and will not,  cause Go team prefers simple manageable compiller, and this inlining is very complex. 

More over: this custom loop avoids oone of boundary checks,  and, in fact, range loop is recommended way to awoid boundary check. 

Measure,  do not suppose.

funny-falcon: https://gist.github.com/funny-falcon/43de34705edf727f9f8b

neurodrone: Thanks for humoring me and writing the bench @funny-falcon! I had ran it through `-gcflags=-m` (where it had mentioned it would inline `sort.Less()`) but realized too late that the it's ability to do that _within_ `sort.IsSorted()` was inhibited by its definition that takes in an interface. No more questions from me here. :)

funny-falcon:
fixed error

freelist.go
```diff
@@ -6,19 +6,21 @@ import (
 	"unsafe"
 )
 
+type ctrue struct{}
```
benbjohnson: This seems unnecessary. Can you simply use `map[pgid]struct{}` instead?

benbjohnson:
Besides the `ctrue` type, it lgtm.

freelist.go
funny-falcon: ok, will do

funny-falcon:
I've added commit: reading freelist on write transaction.
If you fill it is ok, then I can replace `do not read freelist if database opened readonly` 2209ae0  with `lazily open freelist on write transaction`  95bec6c

funny-falcon:
last commit looks to be complex :( I will not not be surprised if you reject it.

rgeronimi:
+1
This, along with no-timeout patch #494, makes cooperative sharing of read-access and readwrite-access to the db between concurrent processes much easier and performant. This is ideal for situations where there is low-concurrence between processes but each process needs super-high-intensity access to the db data when it finally obtains the lock (as opposed to a client-server db architecture which is tuned for high-concurrence low-intensity access to the db data).

page.go
```diff
@@ -133,6 +133,17 @@ type pgids []pgid
 func (s pgids) Len() int           { return len(s) }
 func (s pgids) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (s pgids) Less(i, j int) bool { return s[i] < s[j] }
+func (s pgids) isSorted() bool {
```
josharian: Package sort already has an IsSorted function. No need to add this.

freelist.go
```diff
@@ -171,7 +172,9 @@ func (f *freelist) read(p *page) {
 	copy(f.ids, ids)
 
 	// Make sure they're sorted.
-	sort.Sort(pgids(f.ids))
+	if !pgids(f.ids).isSorted() {
```
josharian: Have you benchmarked to confirm that it is faster on sorted data to first ask whether it is sorted? It is not obvious to me. Many sort algorithms run very quickly on already-sorted data, and when f.ids is large, touching all the data twice may be more expensive.

freelist.go
```diff
@@ -9,16 +9,16 @@ import (
 // freelist represents a list of all pages that are available for allocation.
 // It also tracks pages that have been freed but are still in use by open transactions.
 type freelist struct {
-	ids     []pgid          // all free and available free page ids.
-	pending map[txid][]pgid // mapping of soon-to-be free page ids by tx.
-	cache   map[pgid]bool   // fast lookup of all free and pending page ids.
+	ids     []pgid            // all free and available free page ids.
+	pending map[txid][]pgid   // mapping of soon-to-be free page ids by tx.
+	cache   map[pgid]struct{} // fast lookup of all free and pending page ids.
```
josharian: A benchmark or some other data demonstrating the memory savings would be useful. I'm not a prior convinced that the memory savings here are noticeable in real world use.

db.go
```diff
@@ -397,6 +393,19 @@ func (db *DB) close() error {
 	return nil
 }
 
+func (db *DB) ensureFreelist() *freelist {
```
josharian: Can you explain more how this lazy load helps?

It appears that you're trading

* an eager read (that will happen eventually anyway?)
* a regular pointer load
* straightforward code

for

* a lazy read
* an atomic pointer load
* less obvious code

I'd like to see more detail about the benefits to understand why this is worth doing.

page.go
funny-falcon: sort.IsSorted uses interfaces, so it is slower.

freelist.go
funny-falcon: We are talking not about "many sort algorithms", but about sort.Sort.

db.go
funny-falcon: If you open file only for read, you don't need to read freelist at all ever.

page.go
josharian: How much slower? Slower enough that it is worth adding more code to work around it? Numbers would help a lot here.

freelist.go
josharian: Yes. How does sort.Sort behave on already sorted data? Do you have benchmarks to help understand how much of a win this is?

db.go
josharian: The first commit in this series handles that already, no?

josharian:
To be clear, I'm reviewing this (and the other perf-related PRs) because I am not getting the performance I want out of boltdb, and I'm checking to see whether someone else has already fixed my problems. So I *want* boltdb to be faster...

page.go
funny-falcon: If you want numbers, you may measure. If you find I was wrong, I will appologise.

freelist.go
funny-falcon: Why don't you benchmark instead of asking?
Yes, I did benchmark.

db.go
funny-falcon: Sometimes you want to have single code for both readonly open and readwrite simply cause you don't know will you write or not.

Anyway, I was never confident last commit will be accepted.

freelist.go
josharian: Since you bothered to benchmark, why didn't you include it in the commit? And it's not a question of how much faster it is in isolation, it's a question of how much faster it is in the context of things that matter in boltdb.

Anyway, I'll stop reviewing now.

DavidVorick:
@funny-falcon  Any sort of speedup PR should always have reproducible benchmarks demonstrating the advantages of the new code.

freelist.go
funny-falcon: Look, I was investigating Bolt more than year ago. I need it to open many mostly-read-only files fast. That is why I concentrated on open. All patches were benched, but I have no numbers now. Neither I use Bolt now, cause my director didn't ratify my suggestion of using it.
So now I don't care will this pull request be accepted or not.
If you benchmark it and find it will help you, then it will be your deal to push on merging this pull request.
Though, if you find some error here, I will fix it.

funny-falcon:
@DavidVorick , I don't care will this pull request be accepted, or not.
If you want numbers, measure by your self.
Though, if you find error, I will fix it.

page.go
aybabtme: Usually the onus is on you to prove that your performance claims are justified. Without having provided proofs that your optimizations were guided by a scientific method, it's to be expected that others will ask you to justify your claims.

aybabtme: @funny-falcon thanks for your feedback, I hadn't yet read the other comments. I now see that this is an old PR, sorry for the confusion. We can disagree on the burden of proof another time maybe. Have a good day.

funny-falcon:
Netherless, I've rebased patch and simplified couple of details.

PR #413, opened 2015-09-04
Release read-write Tx after close

benbjohnson:
## Overview

This pull request removes references to the last write transaction and and its dirty pages after the transaction has been closed. This bug did not affect the safety of the data, however, it would cause dirty pages to linger in-memory until the next write transaction began.

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -236,7 +236,8 @@ func (tx *Tx) close() {
 		var freelistPendingN = tx.db.freelist.pending_count()
 		var freelistAlloc = tx.db.freelist.size()
 
-		// Remove writer lock.
+		// Remove transaction ref & writer lock.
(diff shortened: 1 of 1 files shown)


tx.go
```diff
@@ -250,7 +251,12 @@ func (tx *Tx) close() {
 	} else {
 		tx.db.removeTx(tx)
 	}
+
+	// Clear all references.
```
otoolep: Could you put this stuff in a defer? Would it make sense? Perhaps the code is just clearer down here.

benbjohnson: The function doesn't have any other exit points. I'm not sure a `defer` would help.

otoolep: Yeah, makes sense.

PR #414, opened 2015-09-08
add bolt_arm64.go for arm64

jefby:
Support arm64 architecture

Signed-off-by: jefby <jef199006@gmail.com>

### DIFF
--- a/bolt_arm64.go
+++ b/bolt_arm64.go
@@ -1,0 +1,7 @@
+package bolt
+
+// maxMapSize represents the largest mmap size supported by Bolt.
+const maxMapSize = 0xFFFFFFFFFFFF // 256TB
+
+// maxAllocSize is the size used when creating array pointers.
+const maxAllocSize = 0x7FFFFFFF


jefby:
@benbjohnson Hi, could you please check the code? Thanks very much.

benbjohnson:
:+1: thanks!

bolt_arm64.go
```diff
@@ -0,0 +1,7 @@
+package bolt
+
+// maxMapSize represents the largest mmap size supported by Bolt.
+const maxMapSize = 0xFFFFFFFFFFFF // 256TB
+
+// maxAllocSize is the size used when creating array pointers.
+const maxAllocSize = 0x7FFFFFFF
```
drsnyder: This is causing build errors:

```
src/github.com/boltdb/bolt/bolt_arm64.go|4 error| maxMapSize redeclared in this block previous declaration at src/github.com/boltdb/bolt/bolt_amd64.go:4                                                                                                                                                                                                                                                                         
src/github.com/boltdb/bolt/bolt_arm64.go|7 error| maxAllocSize redeclared in this block previous declaration at src/github.com/boltdb/bolt/bolt_amd64.go:7
```

benbjohnson: Are you running Go 1.4.x?

benbjohnson: I'll move this discussion to https://github.com/boltdb/bolt/issues/416.

benbjohnson:
I reverted this change in this PR (https://github.com/boltdb/bolt/pull/418). I attempted to add build tags to support it in this PR (https://github.com/boltdb/bolt/pull/417) but no luck.

@jefby If you can get build tags working on Go 1.4 then we can put arm64 support back into Bolt.

jefby:
@benbjohnson Hi,it can work in go 1.5beta3/1.5/1.5rc1, maybe version 1.4.x is not supported arm64.

benbjohnson:
It works in 1.5 but the problem is that 1.4 doesn't recognize arm64 as a GOARCH so it doesn't ignore the file. See the other PR for more details.

dodgerblue:
Any current plans on merging the arm64 support in bolt? From the discussion, I understood that the build fails for Go 1.4.x. However, I am now trying to build Docker 1.9-dev on arm64 and this patch would be the missing piece. What's the current status?

dodgerblue:
Oh and by the way - Go version is 1.5.

benbjohnson:
@dodgerblue The issue is that Go 1.4 (and before) don't recognize `arm64` as an `ARCH` so builds fail if `bolt_amd64.go` is included. It's a tough choice to either support Go 1.4 or Go 1.5. If you can get the arm64 to be ignored by Go 1.4 then I'm cool to merge it in.

dodgerblue:
@benbjohnson What about #417? Doesn't that fix the issue? I tested that on my machine and got no compile errors (albeit it's not an amd64, but a i386). So if I were to submit a PR, it would be consisted of #414 and #417.

benbjohnson:
@dodgerblue Hmm, yeah, you're right. If you can submit the PR and test it against Go 1.4 & 1.5 then I'll merge it in. :+1:

hqhq:
@benbjohnson I sent a new PR #435 , added build tag for arm64, I tested it against Go 1.4.2 & 1.5, it works fine.
And I think we can add that kind of build tags for other platform specific files, like other Docker venders such as https://github.com/vishvananda/netns does? I'll sent a new PR soon.

hqhq:
I checked other vendors, like https://github.com/kr/pty
it only handles build tag for arm64, so maybe we can keep other files as they were.

PR #419, opened 2015-09-15
windows: implement file locking

raggi:
I added file locking for Windows. Some links to reference constant sources are included.

Tested manually using an example program that simply opens a DB and waits for signal, against Windows XP (sadly, my target platform). According to MSDN this should work on future platforms also, see referenced documentation.

The timeout support was derived from flock_unix.go.

### DIFF
--- a/bolt_windows.go
+++ b/bolt_windows.go
@@ -8,6 +8,38 @@ import (
 	"unsafe"
 )
 
+// LockFileEx code derived from golang build filemutex_windows.go @ v1.5.1
+var (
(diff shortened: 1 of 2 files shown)


raggi:
See #420 for why I wasn't able to run the whole test suite on my target platform.

benbjohnson:
@raggi Thanks for digging into this one. If you can successfully run with the test suite with a lower `MaxValueSize` (as mentioned in #420) then I'll merge this in.

PR #427, opened 2015-09-21
Add two little projects to README project list

joyrexus:
So, [stow](https://github.com/djherbis/stow) is just a little utility to simplify de/encoding values and [buckets](https://github.com/joyrexus/buckets) is a tiny wrapper/extension, but some might find 'em handy for the particular use cases they're targeting.

I considered adding [simplebolt](https://github.com/xyproto/simplebolt) as well, but that might qualify as more of a personal experiment than something designed for other users.  Not sure.  (I guess there are tests and a license.)

Feel free to close/reject this pull request if you'd prefer a separate sub-heading in the README for utilities/extensions/experiments, restricting the current list to notable public/opensource projects?  Or perhaps a project wiki page, organized by broad categories?


benbjohnson:
Subheadings would probably be good but I feel strange making a distinction between notable and non-notable projects. Also, notable projects usually have humble beginnings so it's impossible to know what's going to eventually be notable.

Thanks for the PR!

PR #428, opened 2015-09-21
Document undefined behavior in ForEach

### LINKED ISSUE #426: Deleting items from a bucket during ForEach
If you delete items from a bucket during a ForEach, you can get non-deterministic behavior (the specific problem I ran into was that sometimes not all items would be reached).

Given that go maps work the same way, I probably should have expected this, but it's not documented that you aren't supposed to delete elements from a bucket while iterating through it with ForEach.

lukechampine:
fixes #426


benbjohnson:
:+1: thanks!

PR #432, opened 2015-10-03
introduce initialMmapSize

xiang90:
InitialMmapSize is the initial mmap size of the database in bytes.
Read transaction is guaranteed not to block write transaction
if the mmap size required by the database is smaller than
the initial mmap.


db.go
```diff
@@ -672,6 +676,18 @@ type Options struct {
 	// Open database in read-only mode. Uses flock(..., LOCK_SH |LOCK_NB) to
 	// grab a shared lock (UNIX).
 	ReadOnly bool
+
+	// InitialMmapSize is the initial mmap size of the database
+	// in bytes.
+	//
+	// Read transaction is guaranteed not to block write transaction
```
clutchski: "Read transactions are guaranteed to not block write transactions ..."

db_test.go
```diff
@@ -304,6 +304,47 @@ func TestOpen_ReadOnly(t *testing.T) {
...
+// effect. This is a very hacky test since the mmap size is not exposed.
+//
+// This test indirecly ensures the behavior by trying to block the re-mmap
+// by a reader and aks a writer to write a large value that will trigger
+// a mmap if the initial mmap size is small.
+func TestDB_Open_InitialMmapSize(t *testing.T) {
+	path := tempfile()
+	defer os.Remove(path)
+
+	isz := 1 << 31 // 2GB
```
ngaut: I think zero size is better, so we don't need to commit a large write :)

benbjohnson:
@xiang90 were you able to test this out on etcd?

xiang90:
@benbjohnson We will test it before early next week. Sorry for the delay.

db_test.go
gyuho: *indirectly?

xiang90:
@benbjohnson Yes. We tested it in both etcd and a test case. It works well. See #472 for more information. @gyuho is taking over this.

PR #435, opened 2015-10-09
Add support for arm64

hqhq:
Added build tag for arm64, so we won't get any errors on other
platforms with go 1.4.x.

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>


dodgerblue:
LGTM

benbjohnson:
:+1:

PR #439, opened 2015-10-17
In TestBucket_Delete_FreelistOverflow, use Cursor.Delete to remove records.

MJDSys:
When removing all the keys from the bucket, use Cursor.Delete directly, which
avoids a seek through the bucket to find the element to delete.  This knocks
1/5 of the time off my laptop when running this test while not reducing the
amount of overflow pages cleared out of the free list.


benbjohnson:
:+1:

Issue #324, opened 2015-03-06
bucket.Put() should call cloneBytes() consistently on the Key and Value


cep21:
The documentation inside "Caveats & Limitations" contains 

    Byte slices returned from Bolt are only valid during a transaction

This documents slices used during reads.  It is also true that the slices passed to a bolt transaction are required to be valid for the entire transaction.

I attempted to memory optimize my code by reusing the same byte slice inside a transaction and doing Put() operations with it.  This did not work.  Maybe it would be good to clarify the Caveats & Limitations or the documentation for Put() to mention that the slice passed must be untouched for the duration of the transaction.

cep21:
I looked more into this and it may be a bug.  You call cloneBytes() inside bucket.Put() on the value, but do not on the key.

davecgh:
I wholeheartedly disagree.  In fact, I think it was a mistake for the key to be copied too.  I don't want the ability to avoid additional copies taken away from me as a caller who is in need of efficient use.  The API should document that it has zero-copy semantics and the caller must be careful to not reuse/modify the data.

I understand why it's convenient, but the caller can make a copy if they so desire.  When you shove it into the Put function, the caller no longer has the opportunity to avoid the copy and that is bad for performance.

cep21:
I don't have a big opinion either way.  I also would like it if the bytes were just pointers, but then I need some kind of guarantee about when it's safe to reuse the buffer I've passed in.  Is it safe to reuse when the transaction is over?  Or maybe the contract is that I can't reuse the buffer ever?  It also makes sense to be consistent about how the bytes of the key and value are treated, rather than only calling copy on one.

I hope it doesn't sound like I'm splitting hairs, but contracts on how I can use the []byte I pass into Put() both key and value are important for optimizing my code.

tv42:
I'd vote for less copying, more exact documentation of the assumptions. Transaction scope seems sane.

This does make Put different from e.g. Write. But Get is already different from Read, I think that's ok.

artyom:
I would also like this to be more clearly specified in `Put` function documentation — this would have saved me at least an hour of figuring out what's happening. It would be great if documentation warn against reusing byte slice passed to `Put` so we can avoid [usage attempts like this one](https://gist.github.com/artyom/805ae04bfa4ae2f8df10):

```
	err = db.Update(func(tx *bolt.Tx) error {
		bkt, err := tx.CreateBucketIfNotExists([]byte("one"))
		if err != nil {
			return err
		}
[...]

Issue #429, opened 2015-09-25
Boltdb Deleting Bucket?


DavidVorick:
I've started poking around in the bolt codebase, but it seems like, nondeterministically, though after a significant amount of changes within a single tx, calling `Delete` on a bucket can result in the entire bucket being deleted.

edit: it doesn't always drop the whole bucket. Sometimes calling `delete` will just drop 3/5 items. Maybe it's getting confused about a leaf node vs. root/branch node and dropping a bunch of children?

edit2: inside of `Bucket.Delete`, a call to `c.node().del(key)` is made. Seems like the failure only occurs when the index of the leaf is `0` and len(node.inodes) is `1`. Further testing continues to confirm this trend.

Things I know about my code:

1. A _lot_ is happening in a single tx. Probably on the order of thousands of reads and writes to the bucket.
2. The change that is causing the bucket to be emptied is not happening inside of a ForEach statement or a cursor.
3. The bucket seems most likely to wipe when there are only a few elements remaining. The stress test brings it up as high as 50 items, but it usually fails at 2 or 3 items. ~~I haven't seen it fail with more than 6 items remaining in the bucket.~~ (got it to drop from 18 to 10 items in a single delete operation by increasing the stress)
4. The failure is nondeterministic. Sometimes there won't be a failure, but I can get the failure happening again by adding printlns in arbitrary places in my code (the logic is not changed at all, but it can have an effect on whether the test fails or not)
```
    lenScoBucket(tx) // prints 3
    err := scoBucket.Delete(id[:]) // err is nil
    lenScoBucket(tx) // prints 0
```

My whole codebase is pretty complicated. It's completely possible that I'm doing something else incorrect somewhere but the error isn't showing up until this point. I will be digging through the bolt code, but was hoping someone had ideas.

davecgh:
Is the transaction shared across goroutines?  I ask because it's not safe to do that and you mentioned that you can cause the error by adding prints which changes timing characteristics and therefore smells like a race condition.

If you're deleting different IDs from different goroutines using the same transaction, the type of behavior you describe could certainly happen due to the fact `Delete` involves a seek which sets state in the transaction and a race there could end up with an incorrect node being deleted.

DavidVorick:
@davecgh thanks for dropping in.

The tx is happening as 1 goroutine. The program does spawn hundreds of goroutines but the tx is only active inside of 1 of them. I'll double check that though. Additionally, I have the race detector on, and it's not complaining at all.

[...]

PR #441, opened 2015-10-20
bucket: document Put() value must remain valid for the transaction

pmezard:
Issue #324

I have been bitten by this too. Let's try to move forward on the documentation (or implementation).

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -270,6 +270,7 @@ func (b *Bucket) Get(key []byte) []byte {
 
 // Put sets the value for a key in the bucket.
 // If the key exist then its previous value will be overwritten.
+// Supplied value must remain valid for the life of the transaction.
 // Returns an error if the bucket was created from a read-only transaction, if the key is blank, if the key is too large, or if the value is too large.
 func (b *Bucket) Put(key []byte, value []byte) error {
 	if b.tx.db == nil {


DavidVorick:
What is meant by 'supplied value must remain vaild for the transaction'?

We're experiencing bugs, maybe this is the source of our bugs?

pmezard:
It means the "value" parameter is held by the transaction and will be evaluated at Commit() time. See the bug referenced above for details.

benbjohnson:
I think it's probably worth copying the value byte slice on `Put()`. Then it would be consistent with the key. Also, it's unlikely to have a high overhead compared to the overhead of writing pages to disk.

Can you PR the copy() change instead?

MJDSys:
@benbjohnson I doubt the copy would cost much, but I would worry that this would cause large amounts of garbage to be generated.  If we are going to move to copying both, could the sync.Pool be used?  And/Or create a second function the explicitly does no copy?

I've been playing a test in Bolt, to see about speeding it up, and I noticed that it has a large amount of CPU time dealing with the allocator/GC.  And other programs using local K/V stores often suffer this too.  So this isn't theoretical, and a huge plus to not run into.  Of course safe behaviour is always more important too :)

DavidVorick:
I'm in full support of safe behavior first, performance second. Right now it seems like bolt has a lot of unintuitive and not-well documented 'gotcha's - we've been bitten by a handful. I'm wondering if the issue I have open right now (#429) isn't somehow related to this.

benbjohnson:
@MJDSys I agree that the extra GC sucks. I could _maybe_ see a "NoCopy" flag on the `Bucket` if there's profiling data to support a case for a significant performance gain. A lot of people (myself included) encode/decode to and from the `value` byte slice so that's already doing a copy.

@pmezard @DavidVorick I went ahead and posted a PR for this fix: #443. Can you test it out locally to sanity check it?

MJDSys:
@benbjohnson I have seen profiling data saying that copying data in the database layer is expensive, due to GCs.  The only bolt related application I've seen do this is a test in your suite.  I know syncthing has significant CPU usage for me using leveldb, a large amount of which is spent in the GC.  So if you are doing many writes this can become a problem, and thus why I suggest avoiding it.

I agree a safe by default option is best, so only people who have write loads high enough to cause issues need to worry.  If you are happy to take such a flag in the future, should someone need it, I agree going safe first is best.

pmezard:
Replaced by https://github.com/boltdb/bolt/pull/443

DavidVorick:
@benbjohnson ran the PR on our test suite. Did not fix #429, but also did not cause noticeable slowdown.

pmezard:
Reopening documentation fix after code fix was rejected.

benbjohnson:
:+1:

pmezard:
@benbjohnson maybe close https://github.com/boltdb/bolt/issues/324

benbjohnson:
Thanks!

PR #442, opened 2015-10-20
README: mention the memory map consumes a lot of address space

pmezard:
I have been bitten by this when running a simple application on a
Raspberry Pi 2. It had 3 stores: two 512M stores and one 128M. The
application refused to start the moment the first index grew from 512M
to 1G.

Bolt is really nice but it is easy to overlook this limitation,
especially with the note about supporting stores much larger than
*physical* memory.


benbjohnson:
Good addition. Thanks!

PR #443, opened 2015-10-20
Copy Bucket.Put() value byte slice

benbjohnson:
## Overview

This pull requeszt adds a byte slice copy of the `value` argument to the `Bucket.Put()` function. This is needed because changes to the `value` argument after the function returns can cause unexpected
results.


davecgh:
I've mentioned this previously, but I'm really not a fan of this change.  If I'm putting a large value (on the order of a few MBs which isn't all that uncommin), forcing it to be copied in `Put` means I have to incur another copy on a database that typically prides itself on being zero-copy.

benbjohnson:
@davecgh Bolt's primary aim is to be safe, easy and reliable and this is an issue that's bitten too many people IMO.

Can you do some profiling with your app between this branch and `master`? I'm curious to see the impact for large values. I would think that writing data has significantly more overhead than doing a memory copy -- especially since you'd be limited on how many large writes per second you could do so it shouldn't be a lot of overhead for the GC to manage.

davecgh:
Fair enough, I can completely understand the desire to avoid a common issue that ends up biting a lot of people.

Also, I should mention, it's really not a big deal for us for this to be merged since we maintain a vendored copy, so I can simply not pull this change in, but I did want to mention it as a consideration given I doubt we're the only ones that make use of large values (our app is actually a pretty varied mix of lots of small data in on the order of 32-byte hashes with much larger data on the order of MBs).

You are correct that writing the data has a lot more overhead than the memory copy for sure, but there is also additional overhead of zeroing the newly allocated memory, creating more objects to GC,  etc.  Those all do add up (the number of GC objects is more of a factor with lots of small values and keys than fewer larger ones obviously), so much so in fact, that we've historically had to work quite diligently to control all memory allocations since profiling has shown it is completely dominated by memory allocation and GC.  On the plus side, recent improvements in the Go GC garbage collector make a very noticeable difference in that area.

By the way, thanks for all your work on maintaining this and making it open for everyone to use!

pmezard:
@benbjohnson I have forked your branch and added a test there https://github.com/pmezard/bolt/commit/815ea8c7f06a24ae20b1bbad047d2f579994dc5e , feel free to take.

LGTM otherwise.

pmezard:
Just for cross reference, this is about https://github.com/boltdb/bolt/issues/324

mkobetic:
To be honest I'm a bit concerned about this change as well. I don't understand the argument about most people encoding the value so there's copying anyway. To me that's precisely why doing another copy in Put is just wasteful in most cases. Ironically the motivating issue #324 came about from attempting to optimize by reusing a slice for multiple values. I find that quite unconvincing argument for introducing this kind of overhead, that's arguably unnecessary in most practical scenarios. I agree that it's not about speed of memory copying it's about GC load.

Respectfully, I also think that burden of proof should lie with the proponent of a change, not with rest of us proving that a change is harmful.

DavidVorick:
tested this PR in our test suite. Did not cause significant slowdown, did not fix #429.

Overall I support this change, though I'm not an expert at databases. If this change gets rejected, documentation should be improved.

benbjohnson:
@mkobetic You're right that it's causing significant GC overhead. I ran some larger benchmarking tests and the additional copies are causing some large fluctuations in bench times. I'm closing this PR. :+1:

mkobetic:
Thank you, Ben! I really appreciate the extra effort you put in. :bow:

Issue #313, opened 2015-02-22
Document lifetime of Bucket instances


skybrian:
Based on the Tx() and Writable() methods, it looks like a Bucket is contained within a transaction and shouldn't be cached between transactions. But it would be good to document this explicitly.

benbjohnson:
Yes, that's correct. Can you PR a change for the documentation and I'll merge it in?

PR #444, opened 2015-10-21
bucket: document buckets are valid only during the transaction

pmezard:
Issue #313 

Might be a bit heavy handed, but it does the job.


benbjohnson:
Yeah, that's probably a good thing to add to the doc. :+1:

pmezard:
Thanks. Maybe close the related issue (I cannot)?

benbjohnson:
Ah, yeah, np. :)

PR #446, opened 2015-10-27
Patch holes

lukechampine:
This should resolve the issues we've observed, though I suspect `Last` is susceptible to the same bug that `First` and `Next` are. At any rate, this works well enough to ship.


lukechampine:
whoops, meant to PR this into our own repo. Sorry.

Issue #450, opened 2015-10-31
ForEach stops short


lukechampine:
(same issue as #429, but now reproducible.)

`bucket.ForEach` sometimes fails to iterate over the entire bucket if an element has been deleted during the same transaction. Technically, this is a Cursor issue, not a Bucket issue; `Cursor.first` is placing the Cursor at a bad node, causing `Cursor.keyValue` to return nil and `ForEach` to terminate early. Code in question:

```go
func (c *Cursor) first() {
	for {
		var ref = &c.stack[len(c.stack)-1]
		if ref.isLeaf() {
			break
		}
		var pgid pgid
		if ref.node != nil {
			pgid = ref.node.inodes[ref.index].pgid // <- problem occurs here
		} else {
			pgid = ref.page.branchPageElement(uint16(ref.index)).pgid
		}
		p, n := c.bucket.pageNode(pgid)
		c.stack = append(c.stack, elemRef{page: p, node: n, index: 0})
		// ^this elemRef now has a count of 0
	}
}
```
It seems like `ref.node.inodes` can contain an entry referencing a deleted node. These entries are removed by `node.rebalance`, but rebalancing only happens at the end of a transaction. I tried deleting the entry manually (immediately after `node.del`), but this caused panics elsewhere. Alternatively, instead of deleting from `inodes`, we could skip over the bad index. This is the approach used on our fork: https://github.com/NebulousLabs/bolt/pull/1/files

After much trial and error I was able to create a program+database that reproduces this issue. Strangely, when I copy the contents of the bucket to a fresh DB, the issue disappears. I suspect the DB may have some strange properties that are contributing to the problem.

DB: http://pixeldra.in/u/RcC9Kv
Program: http://pixeldra.in/u/0LYHeh

rhcarvalho:
@benbjohnson this issue can be closed. It was mentioned in the PR above, but GitHub didn't understand "Fixes A, B" :sunglasses:

PR #452, opened 2015-11-06
Skip empty pages during cursor seek

### LINKED ISSUE #429: Boltdb Deleting Bucket?

benbjohnson:
## Overview

This commit fixes an issue where keys are skipped by cursors after deletions occur in a bucket. This occurred because the cursor seeks to the leaf page but does not check if it is empty.

Fixes #429, #450

---

@lukechampine Thanks for digging into this and figuring it out. The fix you made makes sense. The cursor needs to move forward to the next item if an empty page is found. I made a small tweak to the fix inside `next()` so that it doesn't recursively call itself. I also added a test case to check for it. I added you as the commit author, though, for attribution.

Can you take a look and verify that this fix works for you in your project?

### DIFF
--- a/cursor.go
+++ b/cursor.go
@@ -34,6 +34,13 @@ func (c *Cursor) First() (key []byte, value []byte) {
 	p, n := c.bucket.pageNode(c.bucket.root)
 	c.stack = append(c.stack, elemRef{page: p, node: n, index: 0})
 	c.first()
+
+	// If we land on an empty page then move to the next value.
+	// https://github.com/boltdb/bolt/issues/450
+	if c.stack[len(c.stack)-1].count() == 0 {
+		c.next()
+	}
+
 	k, v, flags := c.keyValue()
 	if (flags & uint32(bucketLeafFlag)) != 0 {
 		return k, nil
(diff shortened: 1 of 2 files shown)


DavidVorick:
This PR passes all of our unit tests, which the master branch does not do. I'm doing a much more thorough test right now involving our pre-production system, usually takes a while but I'll report back when that finishes.

lukechampine:
Thanks for getting to the bottom of this. Without a solid understanding of bolt's internals, my fixes were a shot in the dark. It was driving me a little crazy, haha.
I am a little surprised that this hasn't caused issues for anyone else. Maybe using `ForEach` after deletions is a rare use case. Does your new test case replicate the issue deterministically?

benbjohnson:
@DavidVorick Awesome! let me know how the rest of the testing goes.

@lukechampine Your fix was pretty spot on. I think most people don't have a need to delete multiple items and then read them from a bucket in the same transaction. Many times mutation and data retrieval occur in separate transactions.

The test case seems to be deterministic. Writing 1000 keys ensures that it'll cover multiple pages and then deleting a large contiguous block at the beginning ensures that the first leaf page will be empty.

DavidVorick:
@benbjohnson rest of testing has gone well. I appreciate you writing this, seems to have solved a frustrating bug

benbjohnson:
@DavidVorick awesome, thanks for testing it out. Just merged. :+1:

PR #453, opened 2015-11-06
do not grow dbsize agressively

xiang90:
Only grow the database size when the high watermark increases.
We also grows the database size a little bit aggressively to
save a few ftruncates and fsyncs.

I have tested this on various environments. The performance impact
is ignorable with 16MB over allocation. Without over allocation,
the performance might decrease 100% when each Tx.Commit needs a new
page on a very slow disk (seek time dominates the total write).

We will not overallocate 16MB when the database is small since it also
limits by the datasz.

### DIFF
--- a/db.go
+++ b/db.go
@@ -84,6 +84,7 @@ type DB struct {
 	dataref  []byte // mmap'ed readonly, write throws SEGV
 	data     *[maxMapSize]byte
 	datasz   int
+	filesz   int // current on disk file size
 	meta0    *meta
 	meta1    *meta
 	pageSize int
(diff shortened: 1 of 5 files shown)


db.go
```diff
@@ -655,6 +656,37 @@ func (db *DB) allocate(count int) (*page, error) {
...
+
+	// over allocate 16MB to avoid calling Truncate aggressively
+	// for efficiency
+	overAllocation := 16 * 1024 * 1024
+	sz = sz + overAllocation
+
+	// do not over allocate
+	if sz > db.datasz {
+		sz = db.datasz
+	}
```
funny-falcon: What if `sz < db.datasz` ?
Why add `overAllocation` and then discard it ?

gyuho:
Kindly ping @benbjohnson. Following up with https://github.com/boltdb/bolt/pull/472, etcd needs this patch as well. It would be great if you can take a look at this.

Thanks!

db.go
benbjohnson: I moved `overAllocation` to `DB.AllocSize` in a122e1c so it can be configurable.

benbjohnson:
@gyuho @xiang90 Sorry for the delay. It takes me a couple hours to go through Bolt PRs (even simple ones) and it's been hard to find the time. I merged the changes and made some minor changes in a122e1c:

1. Move `overAllocation` to `DB.AllocSize` so that it's configurable.

2. Minor cosmetic clean up.

gyuho:
Thanks!

PR #454, opened 2015-11-06
add WriteToFlag to Tx

xiang90:
For in memory workload, it does not make sense to use
o_direct to copy the file. Adding a option to clear out
o_direct and for other future cases.

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -29,6 +29,16 @@ type Tx struct {
 	pages          map[pgid]*page
 	stats          TxStats
 	commitHandlers []func()
+
+	// WriteFlag specifies the flag for write related methods
+	// like WriteTo.
+	// Tx opens the database file with the specified
+	// flag to copy the data.
+	//
+	// By default, the flag is set to empty for in-memory workload.
+	// To avoid cache trashing for large on disk workload, set this
+	// flag with o_direct.
+	WriteFlag int
 }
 
 // init initializes the transaction.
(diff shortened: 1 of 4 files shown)


tx.go
```diff
@@ -29,6 +29,19 @@ type Tx struct {
...
+	// flag to copy the data.
+	//
+	// By default, the flag is set to o_direct to avoid cache
+	// trashing for large on disk workload.
+	// For in-memory workload, clear o_direct can avoid unnecessary
+	// disk IO.
+	//
+	// If open file fails with the given flag, WriteTo will fail
+	// back to use empty flag.
+	WriteToFlag int
```
benbjohnson: Can you change the name to WriteFlag? The interface is WriterTo but this flag would apply to other write methods if they were added later.

tx.go
```diff
@@ -50,6 +63,7 @@ func (tx *Tx) init(db *DB) {
 		tx.pages = make(map[pgid]*page)
 		tx.meta.txid += txid(1)
 	}
+	tx.WriteToFlag = odirect
```
benbjohnson: Can you remove this default? I know it changes the default behavior but I think it improves the behavior for the vast majority of bolt users who have DBs smaller than RAM.

tx.go
```diff
@@ -274,7 +288,7 @@ func (tx *Tx) Copy(w io.Writer) error {
 func (tx *Tx) WriteTo(w io.Writer) (n int64, err error) {
 	// Attempt to open reader directly.
 	var f *os.File
-	if f, err = os.OpenFile(tx.db.path, os.O_RDONLY|odirect, 0); err != nil {
+	if f, err = os.OpenFile(tx.db.path, os.O_RDONLY|tx.WriteToFlag, 0); err != nil {
 		// Fallback to a regular open if that doesn't work.
```
benbjohnson: Can you remove this fallback since the user setting O_DIRECT would know whether it is supported on their system or not.

xiang90:
@benbjohnson All fixed. PTAL.

benbjohnson:
:+1: I made some doc changes and minor clean up in 6b1bbf0ab40ce12920ca9ad28466107a18f47036 as well.

benbjohnson:
/cc @mkobetic This removes `O_DIRECT` as the default. You probably want to add it back in RDB with:

```go
tx.WriteFlags = syscall.O_DIRECT
```

before it calls to `Tx.WriteTo()` for the backup.

xiang90:
@benbjohnson Thank you!

mkobetic:
Thanks @benbjohnson, I made an issue to remind us when we update.

Issue #308, opened 2015-02-18
database file size not updating?


everdev:
First off, thanks for putting together a great DB.  I've been using it on several projects and like it a lot.

I did notice today that I have two databases with the same number of rows: 30,974
I was amazed to see that their file sizes are exactly the same: 134,217,728 bytes

However, they contain a different number of buckets and different values (although the keys are the same).

I also noticed that after deleting tens of thousands of key/value pairs from the database with the following code, the file size remained unchanged.

```
db.Update(func(tx *bolt.Tx) error {
  b := tx.Bucket([]byte("bucketName"))
  err := b.Delete([]byte("keyToDelete"))
  return err
})
```

I double checked the database and it's only returning the correct quantity of key/value pairs when iterating with a ForEach, so the delete did appear to work.

Before deleting the key/value pairs from each database, I moved them to a backup.  The two backup files for the two databases are different files sizes: 67,108,864 and 16,777,216 bytes

So, it looks like it's writing different volumes, but I expected the file sizes of the original databases to be reduced by those respective amounts.

Maybe this isn't an issue, but I'm just curious as to why the file sizes are remaining the same after the deletes and why two different databases would have the same number of bytes.

Thanks!

benbjohnson:
Good questions. Bolt works by allocating 4KB pages and organizing those into a B+tree. It starts at the beginning of the file and keeps allocating more pages at the end as it needs them. Because Bolt is copy-on-write, when a page is updated, its contents get copied to a new page and the old one is freed. So as pages are freed they can be reused the next time Bolt needs to allocate.

For example, let's look at some sweet ASCII art visualization. When you first create a Bolt database it creates a 1MB file which provides space for some metadata pages (`M`), a page for storing free pages (`F`), and some actual data (`D`). The rest of the pages are unallocated (` `):

```
|M|M|F|D| | | | | |
```

When you update your data page, it'll allocate a new page and copy over old data and make any updates there. It'll then free the old page.

```
|M|M|F| |D| | | | |
```

As you start writing more data (and not just updating existing pages), you start to allocate more pages:

```
|M|M|F| |D| | | | |
        ▼
|M|M|F|D|D| | | | |
        ▼
|M|M|F|D|D|D| | | |
        ▼
|M|M|F|D|D|D|D| | |
        ▼
|M|M|F|D|D|D|D|D| |
        ▼
|M|M|F|D|D|D|D|D|D|
```

[...]

Issue #423, opened 2015-09-15
Question: Shrink BoltDB File?


kylebrandt:
We have a database that has grown to 4G. We have a lot of duplicated data that is going to get removed. Is there anyway to compact the boltdb file once the data is removed?

benbjohnson:
Unfortunately there's not a built-in way to compact a database. You can write a simple program to iterate over buckets from a source database and copy them over to a new database and that'll do the same thing as compacting.

chrsm:
I've found that simply using `tx.WriteTo` to back up certain DBs dramatically reduces the size (while obviously keeping the live data).

I don't have any useful metrics/benchmarks, but I could try to put one together and see if it is an actual benefit.

What kind of duplication and scale are we looking at here? Values that were large are completely removed, shrunk, etc?

cpalone:
@benbjohnson is this a feature you're interested in bolt having?

benbjohnson:
@cpalone It'd be nice to do a recursive copy of every bucket instead of simply using `Tx.WriteTo`. That would eliminate an excessive free pages. You can do a simple version where it happens in one transaction but a more efficient implementation would be to batch it at 1000 key/values at a time. That gets a bit more complicated though.

cpalone:
I can't give you a timeline on this but I'll give the batch implementation a shot!

chnrxn:
I would like to have this feature as well. it will be nice to be able to shrink the DB in-place while it is still in use.

funny-falcon:
Unfortunately, it looks like there is no way to find a path to arbitrary page on a disk quickly.
Without this ability it is O(n^2) to compact DB in-place.
Am I right?

benbjohnson:
@funny-falcon That's correct. Given a page ID there's no way to find it's parents without looking through all the branches. Although branch data tends to be relatively small. You could work your way from the end of the database towards the beginning and reallocate pages and their parents to essentially do a defragmentation. Then truncate the database at the highest unfreed page and adjust the freelist.

However, rewriting the data file to a new data file may very well be faster because of the time it takes to find the parents.

tg:
I'm now in need for shrinking my bolt db file although happy to do this offline. Was thinking of creating a CLI tool for this, but what about adding something like `bolt rewrite` to the existing bolt CLI tool?

benbjohnson:
The goal is to add a `bolt shrink` command to the CLI tool but I haven't had the time so far. Feel free to open a pull request for it though!

[...]

PR #460, opened 2015-11-19
Compact db

vincent-petithory:
This adds a `compact` to the bolt tool. It recursively inspects a bolt database and copies its keys to a new database.
This usually results in a compacted database, especially if there was a lot of deleted keys with large values. See #308 

Thanks for taking a look at this!

### DIFF
--- a/cmd/bolt/main.go
+++ b/cmd/bolt/main.go
@@ -102,6 +102,8 @@ func (m *Main) Run(args ...string) error {
 		return newBenchCommand(m).Run(args[1:]...)
 	case "check":
 		return newCheckCommand(m).Run(args[1:]...)
+	case "compact":
+		return newCompactCommand(m).Run(args[1:]...)
 	case "dump":
 		return newDumpCommand(m).Run(args[1:]...)
 	case "info":
@@ -130,6 +132,7 @@ The commands are:
 
     bench       run synthetic benchmark against bolt
     check       verifies integrity of bolt database
+    compact     copies a bolt database, compacting it in the process
     info        print basic info
     help        print this screen
     pages       print list of pages with their types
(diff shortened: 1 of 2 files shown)


jacek99:
This is moderately useful, but forces downtime as the app has to go down and be restarted pointing to a new DB.

It would be great if this could be done on the current DB file while it is being used, to enable high availability (similar to the way Cassandra does compaction while staying online all the time).

chrsm:
See also #423

mdlayher:
Cool stuff, would love to see this integrated into the tool, at least.

Online compaction would be great, but I certainly think this is a nice feature to have regardless.

tg:
Just run this on my 7GB db file and it was successfully truncated to 2GB. Happy days! A few comments:

1. I think source DB should be open in read only mode, otherwise source DB is being modified (when testing on some 167MB sample DB I actually seen the source DB being inflated to 256MB). Although this is an issue with all `bolt` CLI commands like `check` or `stats`, so maybe should be a separate issue.

2. I would suggest output DB should share the same file permissions as source DB rather then `0666`. We already call `os.Stat(path)` so this as simple as taking `Mode().Perm()` from the result and passing to `bolt.Open`.

3. Would it make sense to have a sensible non-zero default for `-tx-max-size`? I bet this would save a few souls from killing their machines while running compaction on very large DBs without reading help carefully. If someone would like to optimise the process using a single transaction, they would need to do it explicitly.

3. As I mentioned in https://github.com/boltdb/bolt/issues/423#issuecomment-235674616, this compacting approach doesn't copy bucket sequence number, but as far as I can see there is no solution to this using the public interface. I guess most people don't use this functionality anyway, so we should at least add some warning in the help message or maybe somehow be able to detect buckets with non-zero sequence numbers and fail the compaction.

benbjohnson:
Sorry for the ridiculously long delay. I posted a pull request which includes the commits from this PR as well as some additional features for copying buckets sequence. The new pull request is #590.

PR #461, opened 2015-11-23
Document iterating over buckets

mark-rushakoff:
I had to dig into the code a bit to understand how sub-buckets work when iterating with a Cursor, so I added that information to the README.


benbjohnson:
:+1: Thanks Mark!

PR #462, opened 2015-12-05
Put backticks around function name in description

rhcarvalho:
All other forms of `FunctionName()` use backticks for formatting, so adding it around NextSequence() too.


benbjohnson:
:+1: thanks!

PR #464, opened 2015-12-05
Apply some changes suggested by golint

rhcarvalho:
Not changed:

```
$ golint ./...
bucket.go:77:25: exported method Root returns unexported type bolt.pgid, which can be annoying to use
page.go:138:1: receiver name a should be consistent with previous receiver name s for pgids
cmd/bolt/main.go:1354:41: exported func ReadPage returns unexported type *main.page, which can be annoying to use
```

I think the `bucket.Root()` method should probably have a different signature, but it was added over a year ago and I wonder how many people we'd be breaking. It's not used in the current code base.

The receiver name in `page.go:138:1` seems just fine in its context, and `cmd/bolt/main.go:1354:41` is not a big deal as `cmd/bolt` is not meant to be imported (but we could make functions there unexported anyway if wanted).

I'm also not changing tests and benchmark function names that have underscores, but we could as well.

Looking forward for some feedback on these changes :smile:

### DIFF
--- a/batch.go
+++ b/batch.go
@@ -43,7 +43,7 @@ func (db *DB) Batch(fn func(*Tx) error) error {
 	db.batchMu.Unlock()
 
 	err := <-errCh
-	if err == trySolo {
+	if err == errTrySolo {
 		err = db.Update(fn)
 	}
 	return err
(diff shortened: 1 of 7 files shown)


rhcarvalho:
Closing due to age.

PR #469, opened 2015-12-07
Make bolt.Open return the documented errors

### LINKED ISSUE #468: bolt.Open returns non-standard error
In [db.Open](https://github.com/boltdb/bolt/blob/b34b35ea8d06bb9ae69d9a349119252e4c1d8ee0/db.go#L192), instead of returning a standard error from `errors.go`, the validation error is wrapped with `fmt.Errorf`, which makes it harder to write code that handles specific errors.
I see that it is probably to help with debugging, as the validation error can happen in page "meta0" or "meta1".

I'd like to be able to write code like so:

```go
	db, err := db.Open("my.db")
	switch err {
	case bolt.ErrChecksum:
		// handle err...
	case bolt.ErrInvalid:
		// handle err...
	case bolt.ErrVersionMismatch:
		// handle err...
	}
```

That wouldn't be necessary if there was a clear way to check that a file is a valid database. Checking `err.String() == "meta0 error: invalid database"` doesn't look good.

rhcarvalho:
Errors from calling `bolt.Open` will be as documented in `errors.go`:

> **ErrInvalid** is returned when a data file is not a Bolt-formatted database.
> **ErrVersionMismatch** is returned when the data file was created with a different version of Bolt.
> **ErrChecksum** is returned when either meta page checksum does not match.

Fixes #468.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -7,34 +7,120 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"regexp"
 	"runtime"
 	"sort"
 	"strings"
 	"testing"
 	"time"
+	"unsafe"
 
 	"github.com/boltdb/bolt"
 )
 
 var statsFlag = flag.Bool("stats", false, "show performance stats")
 
-// Ensure that opening a database with a bad path returns an error.
-func TestOpen_BadPath(t *testing.T) {
-	db, err := bolt.Open("", 0666, nil)
-	assert(t, err != nil, "err: %s", err)
-	assert(t, db == nil, "")
+// version is the data file format version.
+const version = 2
+
+// magic is the marker value to indicate that a file is a Bolt DB.
+const magic uint32 = 0xED0CDAED
+
+// pageSize is the size of one page in the data file.
+const pageSize = 4096
+
+// pageHeaderSize is the size of a page header.
+const pageHeaderSize = 16
+
+// meta represents a simplified version of a database meta page for testing.
+type meta struct {
+	magic    uint32
+	version  uint32
+	_        uint32
+	_        uint32
+	_        [16]byte
+	_        uint64
+	_        uint64
+	_        uint64
+	checksum uint64
 }
 
 // Ensure that a database can be opened without error.
 func TestOpen(t *testing.T) {
 	path := tempfile()
-	defer os.Remove(path)
 	db, err := bolt.Open(path, 0666, nil)
 	assert(t, db != nil, "")
 	ok(t, err)
 	equals(t, db.Path(), path)
 	ok(t, db.Close())
+}
+
+// Ensure that opening a database with a bad path returns an error.
+func TestOpen_BadPath(t *testing.T) {
+	for _, path := range []string{
+		"",
+		filepath.Join(tempfile(), "youre-not-my-real-parent"),
+	} {
+		t.Logf("path = %q", path)
+		db, err := bolt.Open(path, 0666, nil)
+		assert(t, err != nil, "err: %s", err)
+		equals(t, path, err.(*os.PathError).Path)
+		equals(t, "open", err.(*os.PathError).Op)
+		equals(t, (*bolt.DB)(nil), db)
+	}
+}
+
+// Ensure that opening a file with wrong checksum returns ErrChecksum.
+func TestOpen_ErrChecksum(t *testing.T) {
+	buf := make([]byte, pageSize)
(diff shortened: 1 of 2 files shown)


benbjohnson:
@rhcarvalho Looks good. Can you add test cases for them?

db_test.go
```diff
@@ -37,6 +30,45 @@ func TestOpen(t *testing.T) {
...
+		assert(t, err != nil, "err: %s", err)
+		equals(t, path, err.(*os.PathError).Path)
+		equals(t, "open", err.(*os.PathError).Op)
+		equals(t, (*bolt.DB)(nil), db)
+	}
+}
+
+// Ensure that opening a file with wrong checksum returns ErrChecksum.
+func TestOpen_ErrChecksum(t *testing.T) {
+	t.Skip("pending")
```
rhcarvalho: @benbjohnson do you have in mind a simple way to create invalid files that will cause ErrChecksum and ErrVersionMismatch?

benbjohnson: You can create a meta page like this:

```
buf := make([]byte, 4096)
var meta := (*struct {
	magic    uint32
	version  uint32
	_        uint32
	_        uint32
	_        [16]byte
	_        uint64
	_        uint64
	_        uint64
	checksum uint64
})(unsafe.Pointer(&buf[0]))
```

And then set the fields:

```
// Here's an invalid version
meta.magic = 0xED0CDAED
meta.version = 100
```

```
// Here's an invalid checksum
meta.magic = 0xED0CDAED
meta.version = 2
meta.checksum = 123
```

Then write the `buf` to a file and try to open it.

rhcarvalho: Thanks a lot! I had to adapt it a bit to write the tests, as I needed to account for the page header.

rhcarvalho:
@benbjohnson, tests cases were added.

db_test.go
```diff
@@ -37,6 +50,73 @@ func TestOpen(t *testing.T) {
...
+		assert(t, err != nil, "err: %s", err)
+		equals(t, path, err.(*os.PathError).Path)
+		equals(t, "open", err.(*os.PathError).Op)
+		equals(t, (*bolt.DB)(nil), db)
+	}
+}
+
+// Ensure that opening a file with wrong checksum returns ErrChecksum.
+func TestOpen_ErrChecksum(t *testing.T) {
+	buf := make([]byte, 4096)
```
rhcarvalho: Can we avoid the magical 4096 here and write it in terms of something else or at least give it a "name"?

db_test.go
```diff
@@ -37,6 +50,73 @@ func TestOpen(t *testing.T) {
...
+	meta := (*meta)(unsafe.Pointer(&buf[0]))
+	meta.magic = magic
+	meta.version = version
+	meta.checksum = 123
+
+	path := tempfile()
+	f, err := os.Create(path)
+	equals(t, nil, err)
+	f.WriteAt(buf, 0x10) // 0x10 is the page header offset
```
rhcarvalho: Same here about the 16 bytes offset... I was experimenting with adding a `page` struct but not sure it will clarify or obfuscate the tests :)

rhcarvalho:
BTW, is there anything other than source code where I could read about the internal structure of Bolt? Or maybe some paper or documentation about LMDB if they're similar enough.
I'd like to know more about meta pages (for example) than what is mentioned in the README :)

db_test.go
benbjohnson: Sure, you can call it `pageSize`.

db_test.go
benbjohnson: Yeah, I think making it a constant is the best for now. If more of these kinds of tests are needed then maybe it'd make more sense to extract it out.

benbjohnson:
@rhcarvalho Unfortunately I don't know of any good documentation about the internals of either Bolt or LMDB. It's probably worth putting together a blog post though.

Bolt works as an append-only b+tree (although pages get reclaimed once transactions are no longer need them). I found this write up about AOBTs that's pretty good:

http://www.bzero.se/ldapd/btree.html

rhcarvalho:
@benbjohnson thanks for the link!

I've updated the test cases to use constants, please take a look.

benbjohnson:
@rhcarvalho minor nit but otherwise lgtm. Can you squash this commits into one and I'll merge it in.

rhcarvalho:
:ok_hand:

benbjohnson:
@rhcarvalho thanks! :shipit:

PR #471, opened 2015-12-16
Update min mmap size in godoc

rhcarvalho:
The size changed back in https://github.com/boltdb/bolt/pull/375, while the doc has not been updated.


benbjohnson:
Thanks!

PR #478, opened 2016-01-04
Add documentation to Readme for mobile support (iOS/Android)

### LINKED ISSUE #433: anyone tried to get this running on android ?
it performs well on a raspberry, and it begs the question if it can be compiled for mobile. Would be damn useful

TrevorSStone:
Resolves #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,7 +580,7 @@ func (b *BoltDB) Close() {
 Database logic should be defined as methods on this wrapper struct.
 
 To initialize this struct from the native language (both platforms now sync
-their local storage to the cloud. These snippits disable that functionality for the
+their local storage to the cloud. These snippets disable that functionality for the
 database file):
 ####Android
 ```java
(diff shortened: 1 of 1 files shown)


benbjohnson:
lgtm. Is the `BoltDB` wrapper so that `gomobile` can generate Android/iOS wrapper functions? I haven't used the tool before.

Can you squash the commits and then I'll merge it in.

TrevorSStone:
The BoltDB wrapper is what iOS and Android can bind to. Instead of exposing bolt and all of it's methods to java or objc, the logic should all be done in the go layer. The BoltDB struct would most likely be named something else in a real app and then handle all of the storing and retrieving of data specific to the app.

I did not really explore exposing just the base DB type. Hypothetically this might be possible (and do all bolt interaction from java directly) but gobind does have limited type support, and the example I wrote would be the cleaner implementation.

Thanks for looking at the PR, I'll condense the commits when I get home.

PR #479, opened 2016-01-07
Wrong threadCount passed to TestSimulate_1op_1p

azazeal:
TestSimulate_1op_1p should be passing 1 as the threadCount instead of 100.


benbjohnson:
:+1: , thanks!

PR #480, opened 2016-01-08
*: small optimizaton on spill method

gyuho:
1. bucket.go: convert bytes to string only once at spill
2. node.go: access inode slice only once

Related to coreos/etcd#4111.

We are trying to investigate boltdb commit latency spikes in spill methods.
This might not help a lot but please take a look. Thanks!


gyuho:
```
t.tx.Stats(): {PageCount:40 PageAlloc:163840 CursorCount:379 NodeCount:5 NodeDeref:76 Rebalance:0 RebalanceTime:0 S
plit:34 Spill:39 SpillTime:104.697454ms Write:41 WriteTime:3.69173ms}
took: 108.494099ms

total: 37.324798506s
average: 37.324µs
rate: 26791.8392
minimum latency: 1.94µs
maximum latency: 353.976577ms
```

We are seeing (on SSD machine) SpillTime taking more than 100ms. It's very rare but it happens when it spills a lot. If anybody can help investigate, it would be great. Thanks!

gyuho:
With this I see slightly less spikes. It used to have 3 to 5 out 1M put requests that spill takes more than 100ms. Now I see only once or none, if I am doing it correctly.

xiang90:
@gyuho Can you get a benchmark for this change and compare with the result without this change?

gyuho:
OK I will. Thanks!

gyuho:
@benbjohnson @xiang90 I think the spikes we were seeing is from disk I/O.
Please take a look at https://github.com/coreos/etcd/issues/4111#issuecomment-170172414

gyuho:
I was piping the output to disk, which might have affected I/O spikes. I will try to reproduce the spikes more reliably and if it seems like an issue, I will file a separate issue. Sorry for the noise!

gyuho:
Before:

```
./bolt bench -batch-size 1000 -count 1000000 -value-size 128
# Write	1m8.11901751s	(68.119µs/op)	(14680 op/sec)
# Read	1.028665466s	(44ns/op)	(22727272 op/sec)

./bolt bench -batch-size 1000 -count 1000000 -value-size 128
# Write	1m5.422962845s	(65.422µs/op)	(15285 op/sec)
# Read	1.049033985s	(52ns/op)	(19230769 op/sec)
```

After this patch:

```
./bolt bench -batch-size 1000 -count 1000000 -value-size 128
# Write	1m4.7857369s	(64.785µs/op)	(15435 op/sec)
# Read	1.001878111s	(41ns/op)	(24390243 op/sec)

./bolt bench -batch-size 1000 -count 1000000 -value-size 128
# Write	1m4.173369923s	(64.173µs/op)	(15582 op/sec)
# Read	1.013992277s	(42ns/op)	(23809523 op/sec)
```

There's not much difference in performance.

benbjohnson:
@gyuho Closing since there is not a perf gain. Let me know if you want it kept open.

PR #482, opened 2016-01-08
*: fix test print format

gyuho:


benbjohnson:
:facepalm: thanks for fixing these!

PR #486, opened 2016-01-13
Update README.md

tcyrus:
Changed some Badges to SVG and added missing code blocks.


benbjohnson:
Awesome, thanks for cleaning that up!

PR #494, opened 2016-01-23
Added support for no timeout locks on db files

rgeronimi:
Currently the bolt library supports timeouts for db file locking, as well as infinite timeouts. This patch adds support for no timeouts at all.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -13,12 +13,13 @@ import (
 // flock acquires an advisory lock on a file descriptor.
 func flock(f *os.File, exclusive bool, timeout time.Duration) error {
 	var t time.Time
+	if timeout > 0 {
+		t = time.Now()
+	}
 	for {
 		// If we're beyond our timeout then return an error.
 		// This can only occur after we've attempted a flock once.
-		if t.IsZero() {
-			t = time.Now()
-		} else if timeout > 0 && time.Since(t) > timeout {
+		if timeout > 0 && time.Since(t) > timeout {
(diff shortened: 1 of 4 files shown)


rgeronimi:
Hello is there any problem with this patch, or required updates?
Thanks

benbjohnson:
@rgeronimi Sorry, I'm just terrible at managing PRs. Can you add a test for this change? Otherwise it lgtm.

rgeronimi:
Is ok, I thought there was a code reason :)
I will take some time to add a few tests and I come back to you.
Have a nice Xmas

rgeronimi:
Hello, I freed up the time to add the missing test. Unfortunately I was about to add a variant of the exact piece of code removed through
commit b9eb643780505525197ad791e275fc7ebc208bc8 
(@benbjohnson committed on 27 Dec 2016 )
"The new FCTNL locking does not support multiple locks from the
same process which makes those tests fail. The lock tests have
been removed."

I see 2 alternatives:
- In line with this commit we drop entirely the objective to test lock-related functionality
- We use fork a side process to hold the commit

Unfortunately the second alternative requires an external executable (Fork and not ForkExec is not safe to call from within a Go process, as it forgets about Go runtime-related threads in the forked process) which is, depending on how and when tests are run, not trivial.

Should I continue down that second alternative? If yes, I need a separate Go executable which will need to be go-built before the go-test stage begins. Do you see any simpler approach than that?

Thanks

Issue #488, opened 2016-01-14
Provide helper methods for most common data types


jacek99:
The need for manual conversion to []byte on every call is really annoying boilerplate after a while.

Would it be possible to add helper methods to all the APIs for the most common types such as string, int64, etc

e.g.

CreateBucketString(bucket string)
PutString(key string, value []byte)

etc

Just some API enhancements to avoid the constant []byte cast every time.

peter-edge:
I'm working on auto-generating these, I was going to file an issue about it when done and seeing if there's any interest in adding it mainline: https://github.com/peter-edge/boltutil-go

peter-edge:
@jacek99 it's mostly "done, I did move it to https://github.com/peter-edge/bolttype-go (but original link will work, it's a repo rename). https://go.pedge.io/bolttype works too.

`bolttype.DB` is a complete wrapper for `bolt.DB`, except adds auto-generated functions to/from all pairs of:

[]byte
string
uint16
uint32
uint64
pair(uint16,uint16)
pair(uint32,uint32)
pair(uint64,uint64)
pair(uint16,[]byte)
pair(uint32,[]byte)
pair(uint64,[]byte)
pair([]byte,uint16)
pair([]byte,uint32)
pair([]byte,uint64)

benbjohnson:
@jacek99 I agree that it's annoying to convert to `[]byte` all the time. I think that adding helper functions for the different types will make the API explode and make the `godoc` nearly unusable. The types used are very application dependent so I think keeping them in a separate repo works better and users can `import` or copy them to their application directly.

Personally, I typically use `int` or `uint64` for keys so i'll have helper functions like:

```
func u64tob(v uint64) []byte
func btou64([]byte) uint64
``` 

Then for `values` I typically use protobuf encoding so it ends up being `[]byte` anyway.

peter-edge:
But it would be nice to just have this immediately available for common primitives, no? If the naming is consistent, it doesn't matter too much how the godoc looks, ie (from what I just did):

https://godoc.org/go.pedge.io/bolttype - unusable, no questions there
`GetUint64ToBytes(key uint64) []byte` - consistent, I know what function to call
`ForEachUint64ToBytes(fn func(key uint64, value []byte) error) error` - consistent

I also did pairs for some values, see https://github.com/peter-edge/bolttype-go/blob/master/bolttype_test.go

I'd love to just add this as a subpackage at `github.com/boltdb/bolt/type` if you were open to it.

[...]

PR #496, opened 2016-01-25
Add bolttype package

peter-edge:
Signed-off-by: Peter Edge <peter.edge@gmail.com>

This will almost for sure be rejected, but this is re: https://github.com/boltdb/bolt/issues/488.

I wanted to show the maintainers what this would look like as a package within bolt. This makes it separate for current users, and removes the need for everyone to do the same functions over and over.

See the README and godoc for more details.

Thoughts?

@benbjohnson


benbjohnson:
Closed for reasons listed in https://github.com/boltdb/bolt/issues/488#issuecomment-174543637.

peter-edge:
Ah, no fun :) Oh well, I'll keep it as a separate package, see if anyone wants it other than me heh

PR #501, opened 2016-01-28
Use sync.Pool for small pages in db.allocate

LK4D4:
I used #409 as a base for this patch. I decided to treat only `count == 1` as a small page. Also pool is used only for databases with `pageSize == os.Getpagesize`. I think that it's possible to create dynamic pool for each `pageSize` in system, but it's not worth it.
Benchmark results:
```
benchmark                          old ns/op     new ns/op     delta
BenchmarkDBBatchAutomatic-4        2552625       2485200       -2.64%
BenchmarkDBBatchSingle-4           59632698      50757603      -14.88%
BenchmarkDBBatchManual10x100-4     2564789       2452735       -4.37%

benchmark                          old allocs     new allocs     delta
BenchmarkDBBatchAutomatic-4        10199          10202          +0.03%
BenchmarkDBBatchSingle-4           56642          56653          +0.02%
BenchmarkDBBatchManual10x100-4     5986           5995           +0.15%

benchmark                          old bytes     new bytes     delta
BenchmarkDBBatchAutomatic-4        433587        382462        -11.79%
BenchmarkDBBatchSingle-4           32504533      16308931      -49.83%
BenchmarkDBBatchManual10x100-4     1362370       881765        -35.28%
```

### DIFF
--- a/db.go
+++ b/db.go
@@ -36,6 +36,9 @@ const (
 	DefaultAllocSize         = 16 * 1024 * 1024
 )
 
+// default page size for db is set to the OS page size.
+var defaultPageSize = os.Getpagesize()
+
 // DB represents a collection of buckets persisted to a file on disk.
 // All data access is performed through transactions which can be obtained through the DB.
 // All the functions on DB will return a ErrDatabaseNotOpen if accessed before Open() is called.
@@ -320,7 +323,7 @@ func (db *DB) mmapSize(size int) (int, error) {
 // init creates a new database file and initializes its meta pages.
 func (db *DB) init() error {
 	// Set the page size to the OS page size.
-	db.pageSize = os.Getpagesize()
+	db.pageSize = defaultPageSize
 
 	// Create two meta pages on a buffer.
 	buf := make([]byte, db.pageSize*4)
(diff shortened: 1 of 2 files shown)


funny-falcon:
Closed my #409 in favor of this pull request.

benbjohnson:
Sorry for the delay. I did some testing on this and it looks good. I moved the `sync.Pool` into the `DB` (https://github.com/boltdb/bolt/commit/f5f0f7af77195e5339bff58ea4346aa5491835c1) to avoid a global variable. Also, if users are splitting out `DB`s for security reasons then they probably wouldn't want pages shared between DBs in case there's a bug.

Thanks for the PR!

LK4D4:
@benbjohnson Thanks!

funny-falcon:
Imho, it is better to keep global var, and ensure there is no bug.

Doubdfuly, boltdb will work without zeroing this pages before returning to pool, so bug probability is miserable, it will be always caught by tests.

But if DB were split for performance, then global pool is big gain.

benbjohnson:
@funny-falcon I agree that it's unlikely that there's a bug since we're zeroing out. The `DB` is an isolated unit though and I'd like to keep that isolation. Also, I would think that you would still see a good performance gain if you're running multiple DBs with their own pools.

PR #507, opened 2016-02-05
readme modified to be clear on endianness for key/data

dc0d:
Under _Caveats & Limitations_ a little note added at the end of "_The data structures in the Bolt database are memory mapped so..._" paragraph to state explicitly that the developer should use big endian with boltdb.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,7 +437,7 @@ all the keys in a bucket:
 db.View(func(tx *bolt.Tx) error {
 	// Assume bucket exists and has keys
 	b := tx.Bucket([]byte("MyBucket"))
-	
+
 	b.ForEach(func(k, v []byte) error {
 		fmt.Printf("key=%s, value=%s\n", k, v)
 		return nil
@@ -617,7 +617,7 @@ Boltmobiledemo.BoltDB boltDB = Boltmobiledemo.NewBoltDB(path)
 {
     NSURL* URL= [NSURL fileURLWithPath: filePathString];
     assert([[NSFileManager defaultManager] fileExistsAtPath: [URL path]]);
-    
+
     NSError *error = nil;
     BOOL success = [URL setResourceValue: [NSNumber numberWithBool: YES]
                                   forKey: NSURLIsExcludedFromBackupKey error: &error];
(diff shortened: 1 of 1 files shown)


chrsm:
Why? It's not strictly necessary to use big-endian encoded values with boltdb. As a matter of fact, I have quite a large service using boltdb that doesn't encode values at all, only the keys.

The DB file itself won't be portable either way, so it makes little sense to have an encode step for the values themselves.

benbjohnson:
@chrsm Encoding integer keys as big endian puts them in byte sorted order. I think the verbiage can be clearer though.

README.md
```diff
@@ -737,7 +737,9 @@ Here are a few things to note when evaluating and using Bolt:
 * The data structures in the Bolt database are memory mapped so the data file
   will be endian specific. This means that you cannot copy a Bolt file from a
   little endian machine to a big endian machine and have it work. For most
-  users this is not a concern since most modern CPUs are little endian.
+  users this is not a concern since most modern CPUs are little endian. This
+  does not affect your code and you should always use big endian when converting
+  your keys and data to/from byte slice.
```
benbjohnson: Can you move this to a separate bullet point? I think it's unrelated to the endian encoding of the database. Also, it's not clear that the byte encoding is important for sorting specifically.

Maybe something like:

* If you are using integer keys you should use big endian encoding when converting to byte slices so that the keys are in byte sortable order. This allows cursors to iterate over the keys in the expected order.

README.md
```diff
@@ -437,7 +437,7 @@ all the keys in a bucket:
 db.View(func(tx *bolt.Tx) error {
 	// Assume bucket exists and has keys
 	b := tx.Bucket([]byte("MyBucket"))
-
```
benbjohnson: Can you remove this new line?

README.md
```diff
@@ -617,7 +617,7 @@ Boltmobiledemo.BoltDB boltDB = Boltmobiledemo.NewBoltDB(path)
 {
     NSURL* URL= [NSURL fileURLWithPath: filePathString];
     assert([[NSFileManager defaultManager] fileExistsAtPath: [URL path]]);
-
```
benbjohnson: And this one too?

chrsm:
@benbjohnson Right, as mentioned we encode keys - that's probably worthy of being put into the docs somewhere higher than just the "range scans" section (which isn't immediately obvious anyway).

That being said, I don't see a benefit from the perspective of the _value_ - just for consistency's sake, or am I missing something?

dc0d:
@chrsm You are right. I meant just the keys. But it is necessary to mention this somewhere explicitly - more than just in code comment.

chrsm:
Absolutely :+1:

PR #515, opened 2016-02-19
Use tx.meta during Tx.WriteTo()

### LINKED ISSUE #513: panic in freelist.go
In etcd test suite, we keep on hard killing processes.

Also we are keeping on snapshot by using Tx.Copy.

We found boltdb panics when open right after a hard kill around 1 out of 10000 

here is the stack trace
```
goroutine 1 [running]:
runtime.throw(0x153208, 0x5)
	/usr/local/go/src/runtime/panic.go:527 +0x90 fp=0xc82003bb20 sp=0xc82003bb08
runtime.sigpanic()
	/usr/local/go/src/runtime/sigpanic_unix.go:21 +0x1e4 fp=0xc82003bb70 sp=0xc82003bb20
github.com/boltdb/bolt.(*freelist).read(0xc8200661e0, 0x6496000)
	/Users/xiangli/go/src/github.com/boltdb/bolt/freelist.go:162 +0x27 fp=0xc82003bc10 sp=0xc82003bb70
github.com/boltdb/bolt.Open(0xc8200620c0, 0x51, 0x180, 0x2303a0, 0x51, 0x0, 0x0)
	/Users/xiangli/go/src/github.com/boltdb/bolt/db.go:216 +0x5ee fp=0xc82003bd58 sp=0xc82003bc10
main.main()
```

Here is the db file: https://storage.googleapis.com/failure-archive/agent2/2016-02-14T18%3A34%3A44Z/agent.etcd/member/snap/db

benbjohnson:
## Overview

This commit changes `Tx.WriteTo()` to use the transaction's in-memory meta page instead of copying from the disk. This is needed because the transaction uses the size from its meta page but writes the current meta page on disk which may have allocated additional pages since the transaction started.

Fixes #513

/cc @xiang90

### DIFF
--- a/tx.go
+++ b/tx.go
@@ -297,12 +297,34 @@ func (tx *Tx) WriteTo(w io.Writer) (n int64, err error) {
 	}
 	defer func() { _ = f.Close() }()
 
-	// Copy the meta pages.
-	tx.db.metalock.Lock()
-	n, err = io.CopyN(w, f, int64(tx.db.pageSize*2))
-	tx.db.metalock.Unlock()
+	// Generate a meta page. We use the same page data for both meta pages.
+	buf := make([]byte, tx.db.pageSize)
+	page := (*page)(unsafe.Pointer(&buf[0]))
+	page.flags = metaPageFlag
+	*page.meta() = *tx.meta
+
+	// Write meta 0.
+	page.id = 0
+	page.meta().checksum = page.meta().sum64()
+	nn, err := w.Write(buf)
+	n += int64(nn)
 	if err != nil {
-		return n, fmt.Errorf("meta copy: %s", err)
+		return n, fmt.Errorf("meta 0 copy: %s", err)
+	}
+
+	// Write meta 1 with a lower transaction id.
+	page.id = 1
+	page.meta().txid -= 1
+	page.meta().checksum = page.meta().sum64()
+	nn, err = w.Write(buf)
+	n += int64(nn)
+	if err != nil {
+		return n, fmt.Errorf("meta 1 copy: %s", err)
+	}
+
+	// Move past the meta pages in the file.
+	if _, err := f.Seek(int64(tx.db.pageSize*2), os.SEEK_SET); err != nil {
+		return n, fmt.Errorf("seek: %s", err)
 	}
 
 	// Copy data pages.
--- a/tx_test.go
+++ b/tx_test.go
@@ -570,7 +570,7 @@ func TestTx_CopyFile_Error_Meta(t *testing.T) {
 
 	if err := db.View(func(tx *bolt.Tx) error {
 		return tx.Copy(&failWriter{})
-	}); err == nil || err.Error() != "meta copy: error injected for tests" {
+	}); err == nil || err.Error() != "meta 0 copy: error injected for tests" {
 		t.Fatalf("unexpected error: %v", err)
 	}
 }


xiang90:
@benbjohnson We will redeploy our test infra soon with this patch. And I will report back to you soon! Thanks!

nazwa:
Wanted to add my 3 cents here - I'm getting checksum error when opening the copied file. File size also doesn't match - 64KB original vs 48KB new. 

(this was on an almost empty db, windows 10 x64, go1.6)

gyuho:
@nazwa Could you share your code to reproduce that checksum error?

nazwa:
I'm away until tomorrow evening, will share as soon as I get back. But it was as simple as:

1. Open existing db file
2. WriteTo db2 
3. Open db2 in another app 
4. Error.

xiang90:
@benbjohnson We can confirm the original issue mentioned in #513 is fixed.

benbjohnson:
@xiang90 Awesome!

@nazwa I just bought a crappy, cheap Windows laptop on Amazon so I'll give it a test when it arrives on Wednesday. Windows support in Bolt has been iffy lately and that's primarily because I don't have a machine to test on or run CI on.

benbjohnson:
@nazwa Also, if you can provide a small self-contained Go program that replicates the issue that would be awesome.

nazwa:
@benbjohnson I'm impressed by your dedication!

Here you go, this is enough to crash it:

```go
package main

import (
	"os"

	"github.com/benbjohnson/bolt"
)

func panicOnError(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
	var name string
	if len(os.Args) < 2 {
		name = "test.db"
	} else {
		name = os.Args[1]
	}

	db, err := bolt.Open(name, 0666, nil)
	panicOnError(err)

	panicOnError(db.Update(func(tx *bolt.Tx) error {
		bucket, e := tx.CreateBucketIfNotExists([]byte("text_b"))

		panicOnError(e)
		bucket.Put([]byte("k1"), []byte("this is some data"))
		bucket.Put([]byte("k2"), []byte("this is some data"))
		bucket.Put([]byte("k3"), []byte("this is some data"))
		bucket.Put([]byte("k4"), []byte("this is some data"))

		return nil
	}))

	panicOnError(db.Update(func(tx *bolt.Tx) error {
		return tx.CopyFile("copy.db", 0666)
	}))
}
```
`go version go1.6 windows/amd64`


run it first, it will create a test.db file, then run it again with test.db as first parameter to open the newly created file, you should get a panic.

gyuho:
Kindly ping @benbjohnson 

Any updates on this? We have been extensively testing etcd with this patch for more than a week, and never seen the same panic again, as in https://github.com/coreos/etcd/issues/4527. It's been running in Linux.

But as @nazwa mentioned, I also see the same checksum errors in my Windows machine. It would be great if that can be fixed before our next release.

Thanks again!

benbjohnson:
@gyuho I finally got my Windows laptop and got it set up but I haven't had a chance to debug this issue yet. I should be able to in the next day or two.

gyuho:
Got it. Thanks a lot!

benbjohnson:
I did some testing on Windows last night and fixed some locking & truncation issues. It got too late before I could test the changes against this PR, however, I would certainly welcome any testing to those changes: https://github.com/boltdb/bolt/pull/528

I won't have any time today to test those changes against this PR but those changes may have fixed this PR. @gyuho Can you rebase the `windows` branch onto this PR and see if it works on Windows?

gyuho:
@benbjohnson Just tested this patch with https://github.com/boltdb/bolt/pull/528 in the windows machine and seems like I am still getting the same panic?

```
$ go run a.go
panic: checksum error

goroutine 1 [running]:
panic(0x51de40, 0xc082002300)
        C:/Go/src/runtime/panic.go:464 +0x3f4
main.panicOnError(0x390028, 0xc082002300)
        C:/Users/gyuho/a.go:9 +0x52
main.main()
        C:/Users/gyuho/a.go:33 +0x14b
exit status 2

```

Thanks a lot for the help!

benbjohnson:
@gyuho Thanks for checking it! I'm not sure how Windows `WriteTo()` was working after file locking was added to Bolt. I was hoping the locking issue was just masking as a checksum issue. I'll keep digging into it.

Would you mind testing the #528 PR on its own on Windows?

gyuho:
I see. As you suggest, I just ran https://github.com/boltdb/bolt/pull/528 only without this `WriteTo` patch. And there's no checksum error when it runs only with https://github.com/boltdb/bolt/pull/528.

Thanks a lot!

gyuho:
@benbjohnson Could we get any updates on this?
Is this blocked on windows error?

Actually I am working on windows LockFile for etcd `wal` package as well.
And facing the same locking issue. It would be great if you figure out
the correct way of handling flocks in windows. I will play with your windows
patch more.

Thanks!

benbjohnson:
@gyuho Yes, it's just blocked on the windows checksum error. Using a separate lock file fixes the windows locking issue. It's just the checksum now. I should be able to find some time tomorrow to continue looking at it.

gyuho:
Ok got it. Thanks a lot!

nazwa:
Nice! Checking now

nazwa:
Great, this seems to work! Well done

gyuho:
@benbjohnson Thanks! We have just updated boltdb code and deployed with that patch.

PR #527, opened 2016-02-29
Add bolt_ppc.go to compile on 32-bit PPC platforms.

peteheist:
Adding this file allowed me to compile and run bolt in Linux on my old 32-bit PPC Mac Mini.

```
$ cat /proc/cpuinfo 
processor	: 0
cpu			: 7447A, altivec supported
clock		: 1249.999995MHz
revision		: 1.2 (pvr 8003 0102)
bogomips	: 83.24

total bogomips	: 83.24
timebase		: 41620997
platform		: PowerMac
model		: PowerMac10,1
machine		: PowerMac10,1
motherboard	: PowerMac10,1 MacRISC3 Power Macintosh 
detected as	: 287 (Mac mini)
pmac flags	: 00000010
L2 cache		: 512K unified
pmac-generation	: NewWorld
Memory		: 1024 MB
```

### DIFF
--- a/bolt_ppc.go
+++ b/bolt_ppc.go
@@ -1,0 +1,9 @@
+// +build ppc
+
+package bolt
+
+// maxMapSize represents the largest mmap size supported by Bolt.
+const maxMapSize = 0x7FFFFFFF // 2GB
+
+// maxAllocSize is the size used when creating array pointers.
+const maxAllocSize = 0xFFFFFFF


peteheist:
Not sure how this change could cause a CI build failure in TestDB_Put_VeryLarge, but let me know if I've missed something.

benbjohnson:
@peteheist Thanks for the PPC support. I don't have an old enough Mac to test this change on so I'll trust that it works. :) The build failure is unrelated. We're trying to figure out a Windows issue at the moment.

peteheist:
Ok, thanks for taking it. I can at least confirm that it passes the smoke test. :)

I don't know the consequences of hitting maxMapSize or maxAllocSize or if it will be handled correctly, but I will not be hitting those on this test machine. I only assumed that taking the numbers from bolt_386.go would be correct.

benbjohnson:
@peteheist Yeah, those should be reasonable settings. We can re-evaluate them if anyone hits those limits and wants to try different settings.

PR #531, opened 2016-03-04
new bolt_ppc64.go similar to bolt_ppc64le.go

michelmno:
version 4 of the patch

### DIFF
--- a/bolt_ppc64.go
+++ b/bolt_ppc64.go
@@ -1,0 +1,9 @@
+// +build ppc64
+
+package bolt
+
+// maxMapSize represents the largest mmap size supported by Bolt.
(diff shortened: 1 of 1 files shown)


michelmno:
I do not understand the reason of the above failure on continuous-integration/appveyor/pr 
while my change supposed to impact only ppc64 arch

any idea ?

chrsm:
It's due to the failing Windows build.

benbjohnson:
@michelmno I'm working on the Windows build right now. I'm disabling the Appveyor CI until it's fixed. Thanks for the PR! I don't have any way to test ppc64 so I'll trust that this works for you. :)

PR #540, opened 2016-03-22
Add note about Go RFC3339Nano sortability to README

### LINKED ISSUE #536: Please add note about use of Go's RFC3339Nano format
The README suggests the use of RFC3339 as a sortable time format for BoltDB keys: https://github.com/boltdb/bolt/blob/master/README.md#range-scans

Please add a note explaining that, while RFC3339 format is sortable, Go's RFC3339Nano format does not use a fixed number of digits after the decimal and is therefore not sortable.  This causes some subtle bugs for timestamps which are close together.  I'm hoping adding a note to that effect will prevent others from scanning the docs, seeing the suggestion to use RFC3339 for keys and assuming that RFC3339Nano will work as well.

erock2112:
RFC3339 is sortable, but RFC3339Nano is not, because it does not use a fixed number of digits after the decimal.

Fixes #536

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,8 @@ db.View(func(tx *bolt.Tx) error {
 })
 ```
 
+Note that, while RFC3339 is sortable, the Golang implementation of RFC3339Nano does not use a fixed number of digits after the decimal point and is therefore not sortable.
+
 
 #### ForEach()
 


benbjohnson:
:+1:, awesome thanks!

PR #545, opened 2016-03-25
Set slice capacity

### LINKED ISSUE #544: appending to a result of bucket.Get() segfaults (bad capacity)
Expected: capacity to be the same as len.
The following test fails with
```
append_test.go:31:

	exp: 3

	got: 2147483647
```

```go
package bolt_test

import (
    "testing"

    "github.com/boltdb/bolt"
)

func TestAppend(t *testing.T) {
    db := NewTestDB()
    defer db.Close()
    var bn = []byte("bucket")
    var key = []byte("key")
    var val = []byte("val")
    ok(t, db.Update(func(tx *bolt.Tx) error {
        b, err := tx.CreateBucket(bn)
        if err != nil {
            return err
        }
        return b.Put(key, val)
    }))
    var cp int
    ok(t, db.Update(func(tx *bolt.Tx) error {
        v := tx.Bucket(bn).Get(key)
        tx.Bucket(bn)
        cp = cap(v)
        // Seg faults if uncommented.
        // v = append(v, []byte("123")...)
        return nil
    }))
    equals(t, 3, cp)
} 
```

when `v= append...` line is uncommented, it segaults with
```
signal 0xa code=0x2 addr=0x7bf05c pc=0x5f221]

goroutine 5 [running]:
runtime.throw(0x480f00, 0x5)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/panic.go:530 +0x90 fp=0xc82004ac80 sp=0xc82004ac68
runtime.sigpanic()
	/usr/local/Cellar/go/1.6/libexec/src/runtime/sigpanic_unix.go:21 +0x1e4 fp=0xc82004acd0 sp=0xc82004ac80
runtime.memmove(0x7bf05c, 0xc82004ad18, 0x3)
	/usr/local/Cellar/go/1.6/libexec/src/runtime/memmove_amd64.s:132 +0x381 fp=0xc82004acd8 sp=0xc82004acd0
github.com/boltdb/bolt_test.TestAppend.func2(0xc820063a00, 0x0, 0x0)
	/Users/sasha/go/src/github.com/boltdb/bolt/append_test.go:28 +0x20e fp=0xc82004ada0 sp=0xc82004acd8
github.com/boltdb/bolt.(*DB).Update(0xc820065520, 0xc82004aed0, 0x0, 0x0)
	/Users/sasha/go/src/github.com/boltdb/bolt/db.go:547 +0xb3 fp=0xc82004adf8 sp=0xc82004ada0
github.com/boltdb/bolt_test.TestAppend(0xc820082090)
	/Users/sasha/go/src/github.com/boltdb/bolt/append_test.go:30 +0x2ac fp=0xc82004af68 sp=0xc82004adf8
testing.tRunner(0xc820082090, 0x691da0)
	/usr/local/Cellar/go/1.6/libexec/src/testing/testing.go:473 +0x98 fp=0xc82004afa0 sp=0xc82004af68
runtime.goexit()
	/usr/local/Cellar/go/1.6/libexec/src/runtime/asm_amd64.s:1998 +0x1 fp=0xc82004afa8 sp=0xc82004afa0
created by testing.RunTests
	/usr/local/Cellar/go/1.6/libexec/src/testing/testing.go:582 +0x892
```

I also observed the capacity of b.Get() being equal to 8 for a 3 byte value (not reproduced in this test).

It seems that the problem could be here:

```go
// value returns a byte slice of the node value.
func (n *leafPageElement) value() []byte {
    buf := (*[maxAllocSize]byte)(unsafe.Pointer(n))
    return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize] // <--- here
}
```
https://github.com/boltdb/bolt/blob/51f99c862475898df9773747d3accd05a7ca33c1/page.go#L118

I would be fine with either fixing this or with documenting the behavior clearly (appends discouraged).
It seems there is a similar problem with leafPageElement.key (can be reproduced with ForEach).

benbjohnson:
## Overview

This commit sets the capacity on slices returned from `Bucket.Get()` to match the slice length. Previously the capacity would be the size of the mmap max size.

This does not cause any backwards compatibility issues, however, it does allow users to `append()` to the returned slice since that will cause Go to realloc a new slice on the heap.

Fixes #544

/cc @sasha-s

### DIFF
--- a/page.go
+++ b/page.go
@@ -111,13 +111,13 @@ type leafPageElement struct {
 // key returns a byte slice of the node key.
 func (n *leafPageElement) key() []byte {
 	buf := (*[maxAllocSize]byte)(unsafe.Pointer(n))
-	return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos]))[:n.ksize]
+	return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos]))[:n.ksize:n.ksize]
 }
 
 // value returns a byte slice of the node value.
 func (n *leafPageElement) value() []byte {
 	buf := (*[maxAllocSize]byte)(unsafe.Pointer(n))
-	return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize]
+	return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize:n.vsize]
 }
 
 // PageInfo represents human readable information about a page.
--- a/bucket_test.go
+++ b/bucket_test.go
@@ -80,6 +80,46 @@ func TestBucket_Get_IncompatibleValue(t *testing.T) {
 	}
 }
 
+// Ensure that a slice returned from a bucket has a capacity equal to its length.
+// This also allows slices to be appended to since it will require a realloc by Go.
+//
+// https://github.com/boltdb/bolt/issues/544
+func TestBucket_Get_Capacity(t *testing.T) {
+	db := MustOpenDB()
+	defer db.MustClose()
+
+	// Write key to a bucket.
+	if err := db.Update(func(tx *bolt.Tx) error {
+		b, err := tx.CreateBucket([]byte("bucket"))
+		if err != nil {
+			return err
+		}
+		return b.Put([]byte("key"), []byte("val"))
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	// Retrieve value and attempt to append to it.
+	if err := db.Update(func(tx *bolt.Tx) error {
+		k, v := tx.Bucket([]byte("bucket")).Cursor().First()
+
+		// Verify capacity.
+		if len(k) != cap(k) {
+			t.Fatalf("unexpected key slice capacity: %d", cap(k))
+		} else if len(v) != cap(v) {
+			t.Fatalf("unexpected value slice capacity: %d", cap(v))
+		}
+
+		// Ensure slice can be appended to without a segfault.
+		k = append(k, []byte("123")...)
+		v = append(v, []byte("123")...)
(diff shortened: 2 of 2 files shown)


page.go
```diff
@@ -117,7 +118,9 @@ func (n *leafPageElement) key() []byte {
 // value returns a byte slice of the node value.
 func (n *leafPageElement) value() []byte {
 	buf := (*[maxAllocSize]byte)(unsafe.Pointer(n))
-	return (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize]
+	ret := (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize]
```
carlmjohnson: Couldn't you just set the capacity by doing 

```go
ret := (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize:n.vsize]
```

instead of using reflect?

nuss-justin: Couldn't this just be
```go
ret := (*[maxAllocSize]byte)(unsafe.Pointer(&buf[n.pos+n.ksize]))[:n.vsize:n.vsize]
```

carlmjohnson: Great minds think alike. ;-)

benbjohnson: I always forget about the third slice parameter. :)

Fixed in 2b4a0e28943f2dc71b7697cb86764b05e9919637.

benbjohnson:
@sasha-s I fixed it on keys as well in 2b4a0e28943f2dc71b7697cb86764b05e9919637.

benbjohnson:
@sasha-s I compared `bolt bench` on this branch vs `master` and I don't see any performance difference.,

Issue #549, opened 2016-04-07
windows: .lock files not deleted by db.Close


ChrisHines:
On Windows the .lock file is not deleted after db.Close because `db.path` is set to "" before `funlock` is called.

PR on the way.

PR #550, opened 2016-04-07
Delay clearing db.path until just before db.close returns.

ChrisHines:
The Windows version of funlock needs the db.path to delete the
corresponding .lock file.

Fix for #549.

### DIFF
--- a/db.go
+++ b/db.go
@@ -94,7 +94,7 @@ type DB struct {
 	path     string
 	file     *os.File
 	lockfile *os.File // windows only
-	dataref  []byte // mmap'ed readonly, write throws SEGV
+	dataref  []byte   // mmap'ed readonly, write throws SEGV
 	data     *[maxMapSize]byte
 	datasz   int
 	filesz   int // current on disk file size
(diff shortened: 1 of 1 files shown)


db.go
```diff
@@ -383,11 +383,15 @@ func (db *DB) close() error {
...
+
 	db.opened = false
 
 	db.freelist = nil
-	db.path = ""
+
+	// defer clearing the path because funlock may need it.
+	defer func() {
+		db.path = ""
+	}()
```
benbjohnson: Can you move this down to just before the `return nil` at the end of the function? I'd rather not have a `defer` if it's not necessary.

ChrisHines: Is it OK to leave the field populated if one of the error return paths is taken? What invariant is implied by setting db.path = ""?

benbjohnson: There's not an explicit guarantee about `path` when an error occurs. Honestly, I wish I hadn't reset the `path` at all on `Close()` but that's behavior now. I'm OK with changing the behavior on error to not clear the path.

ChrisHines:
@benbjohnson PR updated, PTAL.

benbjohnson:
Thanks @ChrisHines! :+1:

Issue #548, opened 2016-04-06
Crashes can cause data corruption


cyphar:
This was noticed while testing a Docker bug where Docker was killed by the kernel due to a resource allocation issue and the resultant Bolt database is in a corrupted state. It's a bit hard to test this, and I'm trying to get a nice test case that can be tested just using `bolt`. Examples of the errors include:

```
FATA[0000] Error starting daemon: Error initializing network controller: error obtaining controller instance: failed to get bridge network configurations from store: error while populating kmap: meta0 error: checksum error
```

The issue is clearly that the file wasn't written properly, presumably the mapping wasn't properly synced or something like that. Since we're using mmap'd pages I'm not sure what happens if `bolt` is in the middle of modifying the in-memory state of the database and then it crashes -- I assume the kernel will flush the page cache eventually. This is obviously going to be a problem because the state may not be valid.

In addition, we use `fdatasync` rather than `msync`. But if you look at the `msync(2)` man page it makes clear you should only use it:

> msync()  flushes  changes made to the in-core copy of a file that was mapped into memory using mmap(2) back to the filesystem.  **Without use of this call, there is no guarantee that changes are written back before munmap(2) is called.** [emphasis added]

benbjohnson:
> _we use `fdatasync` rather than `msync`_

Bolt uses a read only memory mmap for reading the data (`syscall.PROT_READ`). All changes go through a file descriptor so that's why `fdatasync()` is used.

> _The issue is clearly that the file wasn't written properly_

The issue looks like one of the meta pages was partially written and bolt is not rolling back to the previous transaction correctly. On the plus side, the data file isn't corrupt. The recovery code just needs to fall back if a checksum error occurs.

cyphar:
>> The issue is clearly that the file wasn't written properly
> The issue looks like one of the meta pages was partially written and bolt is not rolling back to the previous transaction correctly. On the plus side, the data file isn't corrupt. The recovery code just needs to fall back if a checksum error occurs.

Maybe I don't understand how boltdb works internally, but is "partially written" data (meaning the process died while modifying the memory that was to be flushed to disk) not a problem? No matter which part of the data it is? I guess that makes sense if transactions are immutable, but are you sure there's no place where this could go badly (like rewriting references to other transactions or something?

benbjohnson:
> is "partially written" data (meaning the process died while modifying the memory that was to be flushed to disk) not a problem?

During a commit, there are two phases:

[...]

PR #555, opened 2016-04-19
db: fix recovery from unsynced metadata

cyphar:
Bolt stores the two latest transactions' metadata, but previously did
not recover from validation failures in the latest by using the second
latest. Fix this by correctly handling validation failures in db.go, as
well as returning the metadata with highest txid which is also valid in
DB.meta().

Fixes: #548

Signed-off-by: Aleksa Sarai <asarai@suse.de>

/cc @benbjohnson


benbjohnson:
@cyphar Thanks! I added some tests for the change and tweaked some docs in a5aec31dc3d13cbd7c0e6faca7489835b0b7e27a.

sigmonsays:
Is it safe to assume that the tests added will catch a regression if this type of crash (and corruption) occurs again?

DaBlitzStein:
+1

PR #557, opened 2016-05-13
Reduce allocations when traversing a database

philpearl:
We use Bolt to store edges for a graph database. Our database is currently about 5GB. We hit a problem enumerating every edge in the graph. We found Bolt would allocate lots of data in the Go heap: over 4GB, which is enough to kill our process.

When we looked into it the memory was being used by Bolt for Cursors and stacks of elemRefs within Cursors. The cursors were mostly used only briefly, so I tried creating pools of freed Cursors & stacks to see if I could reduce allocations. (I tried using sync.Pool, but rolling my own pools seems more effective)

Here are benchmarks before making the change. BenchmarkDBTraverse is a new benchmark I added for my own situation

```
$ go test -bench . -benchmem
seed: 10802
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000
PASS
Benchmark_FreelistRelease10K-8         10000        160469 ns/op       82501 B/op          5 allocs/op
Benchmark_FreelistRelease100K-8         1000       1672175 ns/op      805262 B/op          5 allocs/op
Benchmark_FreelistRelease1000K-8         300       4677185 ns/op     8049008 B/op          5 allocs/op
Benchmark_FreelistRelease10000K-8         30      39605709 ns/op    80413040 B/op          5 allocs/op
BenchmarkDBBatchAutomatic-8              500       2918605 ns/op      429473 B/op      10178 allocs/op
BenchmarkDBBatchSingle-8                  10     167128690 ns/op    16855828 B/op      57609 allocs/op
BenchmarkDBBatchManual10x100-8           300       4281479 ns/op      907296 B/op       6006 allocs/op
BenchmarkDBTraverse-8                1000000          1768 ns/op         272 B/op          5 allocs/op
ok      github.com/boltdb/bolt  170.999s
```

Here are benchmarks from after making the change.

```
$ go test -bench . -benchmem
seed: 81596
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000
PASS
Benchmark_FreelistRelease10K-8         10000        193359 ns/op       82501 B/op          5 allocs/op
Benchmark_FreelistRelease100K-8         1000       1784276 ns/op      805266 B/op          5 allocs/op
Benchmark_FreelistRelease1000K-8         300       4970484 ns/op     8049008 B/op          5 allocs/op
Benchmark_FreelistRelease10000K-8         30      41784659 ns/op    80413040 B/op          5 allocs/op
BenchmarkDBBatchAutomatic-8              500       2741149 ns/op      347806 B/op       8174 allocs/op
BenchmarkDBBatchSingle-8                  10     165493580 ns/op    16899824 B/op      53639 allocs/op
BenchmarkDBBatchManual10x100-8           300       4216260 ns/op      838531 B/op       3986 allocs/op
BenchmarkDBTraverse-8                1000000          1503 ns/op          64 B/op          1 allocs/op
ok      github.com/boltdb/bolt  154.254s
```

Reducing the number of allocations from 5 to 1 in my case is very valuable to me - I should be able to enumerate my database in < 1GB. I expect with only a little more work I can extend the approach to buckets and reduce the allocations to zero per operation.
- Are you interested in taking this change? 
- Is there anything I can do to help you accept it?

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -91,10 +91,7 @@ func (b *Bucket) Cursor() *Cursor {
 	b.tx.stats.CursorCount++
 
 	// Allocate and return a cursor.
-	return &Cursor{
-		bucket: b,
-		stack:  make([]elemRef, 0),
-	}
+	return newCursor(b)
 }
 
 // Bucket retrieves a nested bucket by name.
@@ -110,6 +107,7 @@ func (b *Bucket) Bucket(name []byte) *Bucket {
 	// Move cursor to key.
 	c := b.Cursor()
 	k, v, flags := c.seek(name)
+	c.Close()
 
 	// Return nil if the key doesn't exist or it is not a bucket.
 	if !bytes.Equal(name, k) || (flags&bucketLeafFlag) == 0 {
@@ -162,6 +160,7 @@ func (b *Bucket) CreateBucket(key []byte) (*Bucket, error) {
 	// Move cursor to correct position.
 	c := b.Cursor()
 	k, _, flags := c.seek(key)
+	defer c.Close()
 
 	// Return an error if there is an existing key.
 	if bytes.Equal(key, k) {
@@ -257,7 +256,9 @@ func (b *Bucket) DeleteBucket(key []byte) error {
 // Returns a nil value if the key does not exist or if the key is a nested bucket.
 // The returned value is only valid for the life of the transaction.
 func (b *Bucket) Get(key []byte) []byte {
-	k, v, flags := b.Cursor().seek(key)
+	c := b.Cursor()
+	k, v, flags := c.seek(key)
+	c.Close()
 
 	// Return nil if this is a bucket.
 	if (flags & bucketLeafFlag) != 0 {
@@ -362,6 +363,7 @@ func (b *Bucket) ForEach(fn func(k, v []byte) error) error {
 			return err
 		}
 	}
+	c.Close()
 	return nil
 }
 
--- a/bucket.go
+++ b/bucket.go
@@ -216,6 +216,7 @@ func (b *Bucket) DeleteBucket(key []byte) error {
 	// Move cursor to correct position.
 	c := b.Cursor()
 	k, _, flags := c.seek(key)
+	defer c.Close()
 
 	// Return an error if bucket doesn't exist or is not a bucket.
 	if !bytes.Equal(key, k) {
@@ -291,6 +292,7 @@ func (b *Bucket) Put(key []byte, value []byte) error {
 
 	// Move cursor to correct position.
 	c := b.Cursor()
+	defer c.Close()
 	k, _, flags := c.seek(key)
 
 	// Return an error if there is an existing key with a bucket value.
@@ -317,6 +319,7 @@ func (b *Bucket) Delete(key []byte) error {
 
 	// Move cursor to correct position.
 	c := b.Cursor()
+	defer c.Close()
 	_, _, flags := c.seek(key)
 
 	// Return an error if there is already existing bucket value.
@@ -358,12 +361,12 @@ func (b *Bucket) ForEach(fn func(k, v []byte) error) error {
 		return ErrTxClosed
 	}
 	c := b.Cursor()
+	defer c.Close()
 	for k, v := c.First(); k != nil; k, v = c.Next() {
 		if err := fn(k, v); err != nil {
 			return err
 		}
 	}
-	c.Close()
 	return nil
 }
 
@@ -532,6 +535,7 @@ func (b *Bucket) spill() error {
 			panic(fmt.Sprintf("unexpected bucket header flag: %x", flags))
 		}
 		c.node().put([]byte(name), []byte(name), value, 0, bucketLeafFlag)
+		c.Close()
 	}
 
 	// Ignore if there's not a materialized root node.
--- a/bucket.go
+++ b/bucket.go
@@ -3,6 +3,7 @@ package bolt
 import (
 	"bytes"
 	"fmt"
+	"sync"
 	"unsafe"
 )
 
@@ -126,13 +127,14 @@ func (b *Bucket) Bucket(name []byte) *Bucket {
 // Helper method that re-interprets a sub-bucket value
 // from a parent into a Bucket
 func (b *Bucket) openBucket(value []byte) *Bucket {
-	var child = newBucket(b.tx)
+	child := bucketPool.Get()
+	*child = newBucket(b.tx)
 
 	// If this is a writable transaction then we need to copy the bucket entry.
 	// Read-only transactions can point directly at the mmap entry.
 	if b.tx.writable {
 		child.bucket = &bucket{}
-		*child.bucket = *(*bucket)(unsafe.Pointer(&value[0]))
+		*(child.bucket) = *(*bucket)(unsafe.Pointer(&value[0]))
 	} else {
 		child.bucket = (*bucket)(unsafe.Pointer(&value[0]))
 	}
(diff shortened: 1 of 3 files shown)


philpearl:
Improving the other test cases was just a case of writing the change properly. Here's the latest benchmarks
```
$ go test -bench . -benchmem
seed: 81596
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000
PASS
Benchmark_FreelistRelease10K-8   	   10000	    193359 ns/op	   82501 B/op	       5 allocs/op
Benchmark_FreelistRelease100K-8  	    1000	   1784276 ns/op	  805266 B/op	       5 allocs/op
Benchmark_FreelistRelease1000K-8 	     300	   4970484 ns/op	 8049008 B/op	       5 allocs/op
Benchmark_FreelistRelease10000K-8	      30	  41784659 ns/op	80413040 B/op	       5 allocs/op
BenchmarkDBBatchAutomatic-8      	     500	   2741149 ns/op	  347806 B/op	    8174 allocs/op
BenchmarkDBBatchSingle-8         	      10	 165493580 ns/op	16899824 B/op	   53639 allocs/op
BenchmarkDBBatchManual10x100-8   	     300	   4216260 ns/op	  838531 B/op	    3986 allocs/op
BenchmarkDBTraverse-8            	 1000000	      1503 ns/op	      64 B/op	       1 allocs/op
ok  	github.com/boltdb/bolt	154.254s
```

philpearl:
Added free pools for buckets. This only really helps for read transactions, and only if the application cooperates and says when it is done with the bucket. In my case I now have an average of 0 allocations per operation when scanning my database - a huge win.

```
$ go test -bench . -benchmem
seed: 30215
quick settings: count=5, items=1000, ksize=1024, vsize=1024
00010000000000000000000000000000
PASS
Benchmark_FreelistRelease10K-8   	   10000	    176919 ns/op	   82501 B/op	       5 allocs/op
Benchmark_FreelistRelease100K-8  	    1000	   2129799 ns/op	  805260 B/op	       5 allocs/op
Benchmark_FreelistRelease1000K-8 	     300	   5120436 ns/op	 8049008 B/op	       5 allocs/op
Benchmark_FreelistRelease10000K-8	      30	  46394099 ns/op	80413040 B/op	       5 allocs/op
BenchmarkDBBatchAutomatic-8      	     500	   2983732 ns/op	  350524 B/op	    8213 allocs/op
BenchmarkDBBatchSingle-8         	      10	 168321261 ns/op	16700965 B/op	   53583 allocs/op
BenchmarkDBBatchManual10x100-8   	     300	   4347885 ns/op	  835096 B/op	    3987 allocs/op
BenchmarkDBTraverse-8            	 1000000	      1471 ns/op	       0 B/op	       0 allocs/op
ok  	github.com/boltdb/bolt	160.963s
```

bucket.go
```diff
@@ -144,9 +144,55 @@ func (b *Bucket) openBucket(value []byte) *Bucket {
...
+func (fs *freeBuckets) Put(i *Bucket) {
+	// Ensure bucket is cleaned before storage - otherwise we might stop things
+	// being cleaned up with GC
+	*i = Bucket{}
+	fs.Lock()
+	fs.items = append(fs.items, i)
+	fs.Unlock()
+}
+
+var bucketPool = freeBuckets{}
```
nuss-justin: Is there are reason this doesn't use a sync.Pool? sync.Pool allows the values to still be gc'ed and could even be faster (I haven't testet this). Also sync.Pool has some special optimizations, so it will probably be faster when using multiple goroutines concurrently.

cursor.go
```diff
@@ -20,6 +21,75 @@ type Cursor struct {
...
+}
+
+func (fs *freeStack) Put(e []elemRef) {
+	e = e[:0]
+	fs.Lock()
+	fs.items = append(fs.items, e)
+	fs.Unlock()
+}
+
+var stackPool = freeStack{}
```
nuss-justin: Same question as for the bucketPool.

cursor.go
```diff
@@ -20,6 +21,75 @@ type Cursor struct {
...
+}
+
+func (fs *freeCursors) Put(c *Cursor) {
+	c.bucket = nil
+	fs.Lock()
+	fs.items = append(fs.items, c)
+	fs.Unlock()
+}
+
+var cursorPool = freeCursors{}
```
nuss-justin: Same question as for the bucketPool.

benbjohnson:
@philpearl Just to give you an update, I've been testing out the code and it seems to work fine but I'm concerned about the overhead of locks within `sync.Pool` for users who don't utilize `Close()`. 

Since `Bucket` objects are not safe for concurrent use, I wonder if a simpler approach would be to have an embedded `Cursor` & `[]elemRef` inside `Bucket` that it can use for all its internal operations (`Get()`, Put()`, etc) and only incur one `Bucket` allocation instead of an allocation on every key access.

That doesn't solve the `Bucket` allocation issue though. I think that's a less frequent issue though. One option would be to optionally attach the `bucketPool` to the `Tx`. Otherwise it would always return new allocations.

josharian:
What's the status here? Embedding definitely sounds like a better approach than sync.Pool in this case.

nuss-justin:
I just ran the benchmarks with an embedded cursor in bucket for internal use, to see how much improves the performance.

See [this gist](https://gist.github.com/nuss-justin/b8a57b2370ae6504ebfbed32bb8eb274) for the patch. I haven't really looked at the code so it's probably missing something, but it passes all tests with and without race detector.

This is what I got:

```
name                      old time/op    new time/op    delta
_FreelistRelease10K-8       16.7µs ± 1%    16.5µs ± 1%   -1.21%  (p=0.000 n=10+10)
_FreelistRelease100K-8       213µs ± 1%     212µs ± 1%   -0.31%  (p=0.034 n=10+10)
_FreelistRelease1000K-8     3.30ms ± 0%    3.28ms ± 1%   -0.65%  (p=0.028 n=9+10)
_FreelistRelease10000K-8    31.2ms ± 1%    30.4ms ± 1%   -2.65%  (p=0.000 n=9+9)
DBBatchAutomatic-8          1.71ms ± 1%    1.48ms ± 0%  -13.07%  (p=0.000 n=9+10)
DBBatchSingle-8             27.4ms ± 1%    27.4ms ± 0%     ~     (p=0.280 n=10+10)
DBBatchManual10x100-8       1.67ms ± 1%    1.48ms ± 1%  -11.57%  (p=0.000 n=9+10)

name                      old alloc/op   new alloc/op   delta
_FreelistRelease10K-8       82.5kB ± 0%    82.5kB ± 0%     ~     (all equal)
_FreelistRelease100K-8       805kB ± 0%     805kB ± 0%     ~     (all equal)
_FreelistRelease1000K-8     8.05MB ± 0%    8.05MB ± 0%     ~     (all equal)
_FreelistRelease10000K-8    80.4MB ± 0%    80.4MB ± 0%     ~     (all equal)
DBBatchAutomatic-8           384kB ± 0%     302kB ± 0%  -21.51%  (p=0.000 n=9+10)
DBBatchSingle-8             16.2MB ± 2%    16.2MB ± 1%     ~     (p=0.481 n=10+10)
DBBatchManual10x100-8        901kB ± 1%     819kB ± 2%   -9.09%  (p=0.000 n=10+10)

name                      old allocs/op  new allocs/op  delta
_FreelistRelease10K-8         5.00 ± 0%      5.00 ± 0%     ~     (all equal)
_FreelistRelease100K-8        5.00 ± 0%      5.00 ± 0%     ~     (all equal)
_FreelistRelease1000K-8       5.00 ± 0%      5.00 ± 0%     ~     (all equal)
_FreelistRelease10000K-8      5.00 ± 0%      5.00 ± 0%     ~     (all equal)
DBBatchAutomatic-8           10.2k ± 0%      8.2k ± 0%  -19.77%  (p=0.000 n=10+10)
DBBatchSingle-8              58.8k ± 0%     57.8k ± 0%   -1.69%  (p=0.000 n=10+9)
DBBatchManual10x100-8        6.03k ± 1%     4.04k ± 1%  -33.00%  (p=0.000 n=10+10)
```

go version devel +61db2e4efa Fri Dec 16 17:03:25 2016 +0000 linux/amd64

josharian:
@nuss-justin I don't know the codebase well enough to review. However, I suspect that you want to zero the embedded Cursor (or at least some of its fields) when it is not needed, to allow other things to be GC'd as needed.

Hopefully @benbjohnson will have time to review and get something along these lines merged. :)

josharian:
@philpearl any color on why you closed this?

philpearl:
I raised it a year ago tomorrow. I'm no longer interested in merging it, and it was just untidy in my list of PRs

Incidentally, I've moved my application away from boltdb. I ran into the issues with write-amplification warned about in the readme. I've replaced bolt with in-memory structures plus an append only transaction log.

PR #561, opened 2016-05-18
task#560 print leaf k/v with right value

### LINKED ISSUE #560: PageCommand print leaf k/v with empty value
PrintLeaf get k/v from database, but always get empty value

zhujun2006:
fix #560


benbjohnson:
@zhujun2006 Thanks! Sorry for the delay.

PR #569, opened 2016-06-16
Remove skydb of the README

ifraixedes:
I think that SkyDB is over, I could find any link to the project.


benbjohnson:
Yep, skydb is shutdown. Thanks for removing!

PR #573, opened 2016-07-17
added a BoltDB related project

evnix:
I have added a BoltDB related project.  https://github.com/evnix/boltdbweb
I built it for my personal use as it becomes really easy to experiment with BoltDB.
I think this will be useful for others as well :)


benbjohnson:
Thanks!

Issue #430, opened 2015-09-28
BoltDB panic: runtime error: index out of range, portability issues (ARMv5) - unsafe.Pointer


tfalencar:
Hello.

In a [attempt to get Influxdb running in armv5] (https://github.com/influxdb/influxdb/issues/4218), I bumped with the following problem:

2015/09/24 11:47:10 InfluxDB starting, version 0.9, branch unknown, commit unknown
2015/09/24 11:47:10 Go version go1.5, GOMAXPROCS set to 1
2015/09/24 11:47:10 no configuration provided, using default settings
[metastore] 2015/09/24 11:47:10 Using data dir: /root/.influxdb/meta
**panic: runtime error: index out of range**
goroutine 1 [running]:
**github.com/boltdb/bolt.(Bucket)**.pageNode(0x10cfde80, 0x73676f00, 0x0, 0x0, 0x0)
/root/gocodez/src/github.com/boltdb/bolt/bucket.go:693 +0x2f8
github.com/boltdb/bolt.(Cursor).Last(0x10c3f6ac, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
/root/gocodez/src/github.com/boltdb/bolt/cursor.go:51 +0xb4
[...]

~/gocodez/bin$ uname -a
8 Thu Jun 25 15:31:05 CEST 2015 armv5tejl GNU/Linux
go version go1.5 linux/arm

I've never programmed in go so please excuse me if I'm saying something wrong, but a little bit of investigation makes me think that the usage of the "unsafe" package in multiple locations in BoltDB is causing issues during runtime for different platforms, see:

From [documentation] (https://golang.org/pkg/unsafe/): 
"Package unsafe contains operations that step around the type safety of Go programs.
Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines."
[...]
"Pointer therefore allows a program to defeat the type system and read and write arbitrary memory. It should be used with extreme care."

Now, my "shot in the dark": I see that in BoltDB there are some hardcoded values for "maxAllocSize". Could we maybe change these hardcoded values to unsafe's "Sizeof" function to be more portable?

thiago@debian64VM:~/gocodez/src/github.com/boltdb/bolt$ grep -rnw . -e "maxAllocSize" 
./node.go:205:	b := (*[maxAllocSize]byte)(unsafe.Pointer(&p.ptr))[n.pageElementSize()*len(n.inodes):]
[...]

Issue #327, opened 2015-03-08
TestBucket_Get_NonExistent fail on ARM


yvesf:
```
$ go test -v
seed: 23210
quick settings: count=5, items=1000, ksize=1024, vsize=1024
...
=== RUN TestBucket_Get_NonExistent
panic: runtime error: index out of range

goroutine 20 [running]:
github.com/boltdb/bolt.(*Tx).page(0x10432080, 0x73746500, 0x0, 0xfffffef)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:474 +0xd8
github.com/boltdb/bolt.(*Tx).forEachPage(0x10432080, 0x73746500, 0x0, 0x0, 0x10415e54)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:479 +0x34
github.com/boltdb/bolt.(*Tx).checkBucket(0x10432080, 0x104385e0, 0x104385c0, 0x10438540, 0x1045e0c0)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:379 +0xac
github.com/boltdb/bolt.func·015(0xb6d44020, 0x7, 0xfffffef, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:384 +0xb0
github.com/boltdb/bolt.(*Bucket).ForEach(0x1043208c, 0x10415f10, 0x0, 0x0)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/bucket.go:355 +0x190
github.com/boltdb/bolt.(*Tx).checkBucket(0x10432080, 0x1043208c, 0x104385c0, 0x10438540, 0x1045e0c0)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:387 +0xf4
github.com/boltdb/bolt.(*Tx).check(0x10432080, 0x1045e0c0)
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:338 +0x418
created by github.com/boltdb/bolt.(*Tx).Check
        /home/ebus/influx/gopath/src/github.com/boltdb/bolt/tx.go:315 +0x68
...
```

System information:

```
$ git describe --always
cf33c9e
$ go version
go version go1.4.1 linux/arm
$ uname -a
[...]

PR #578, opened 2016-07-28
Correct broken unaligned load/store in armv5

lorenzo-stoakes:
## Description

armv5 devices and older (i.e. <= arm9 generation) require addresses that are stored to and loaded from to to be 4-byte aligned.

If this is not the case the lower 2 bits of the address are cleared and the load is performed in an unexpected order, including up to 3 bytes of data located prior to the address.

Inlined buckets are stored after their key in a page and since there is no guarantee that the key will be of a length that is a multiple of 4, it is possible for unaligned load/stores to occur when they are cast back to bucket and page pointer types.

The fix adds a new field to track whether the current architecture exhibits this issue, sets it on module load for ARM architectures, and then on bucket open, if this field is set and the address is unaligned, a byte-by-byte copy of the inlined bucket is performed.

Ref: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.faqs/ka15414.html

## Testing

I've tested this fix on actual hardware ([TS-7700](https://www.embeddedarm.com/products/TS-7700)) and run the full suite of tests there (albeit having to reduce some constants due to its limited RAM) - all passing.

## Index out of range

Issues #327 and #430 may well be fixed by this, as the problem usually exhibits itself as an index out of range error.

This occurs because `openBucket()` dereferences a pointer a `bucket` that might not be aligned. Taking a look at the type:

```go
type bucket struct {
	root     pgid   // page id of the bucket's root-level page
	sequence uint64 // monotonically incrementing, used by NextSequence()
}
```

The first field here is a page ID, which is a uint64 (declared as `type pgid uint64`.) When the unaligned access happens it pulls in garbage bytes from up to 3 bytes previous to the address and due to the odd rearrangement of the bytes this number tends to be very large.

Then, when the `pgid` is used to try to reference an address in memory (typically I've found this has occurred in `page()` in `db.go`), `pgid * db.pageSize` exceeds `maxMapSize` and hence the issue manifests as an out of range error:

```go
type DB struct {
	...
	data     *[maxMapSize]byte
	...
}

...

// page retrieves a page reference from the mmap based on the current page size.
func (db *DB) page(id pgid) *page {
	pos := id * pgid(db.pageSize)
	return (*page)(unsafe.Pointer(&db.data[pos]))
}
```

The issue has consistently manifested this way for me prior to the fix.

See this [go playground example](https://play.golang.org/p/JVOtlyxvUY) for a repro + example of the reordering.

### DIFF
--- a/bolt_386.go
+++ b/bolt_386.go
@@ -5,3 +5,6 @@ const maxMapSize = 0x7FFFFFFF // 2GB
 
 // maxAllocSize is the size used when creating array pointers.
 const maxAllocSize = 0xFFFFFFF
+
+// Are unaligned load/stores broken on this arch?
+var brokenUnaligned = false
--- a/bolt_amd64.go
+++ b/bolt_amd64.go
@@ -5,3 +5,6 @@ const maxMapSize = 0xFFFFFFFFFFFF // 256TB
 
 // maxAllocSize is the size used when creating array pointers.
 const maxAllocSize = 0x7FFFFFFF
+
+// Are unaligned load/stores broken on this arch?
+var brokenUnaligned = false
--- a/bolt_arm.go
+++ b/bolt_arm.go
@@ -1,7 +1,28 @@
 package bolt
 
+import "unsafe"
+
 // maxMapSize represents the largest mmap size supported by Bolt.
 const maxMapSize = 0x7FFFFFFF // 2GB
 
 // maxAllocSize is the size used when creating array pointers.
 const maxAllocSize = 0xFFFFFFF
+
+// Are unaligned load/stores broken on this arch?
+var brokenUnaligned bool
+
+func init() {
+	// Simple check to see whether this arch handles unaligned load/stores
+	// correctly.
+
+	// ARM9 and older devices require load/stores to be from/to aligned
+	// addresses. If not, the lower 2 bits are cleared and that address is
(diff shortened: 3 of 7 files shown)


lorenzo-stoakes:
hey @benbjohnson wondered if you had any thoughts on this? Any insight would be very much appreciated! :)

benbjohnson:
@lorenzo-stoakes Sorry for the delay. I'm carving out time each week to do more bolt support. I will look at the change tomorrow and get back with you.

How is the change in https://github.com/resin-os/meta-resin/pull/228 working?

lorenzo-stoakes:
Hey @benbjohnson no problem, thanks very much for taking a look! The fix we merged in meta-resin seems to have resolved the issues we encountered with docker on armv5 altogether so it's looking good in practice for us!

Let me know if you have any feedback/questions/etc. about this patch!

benbjohnson:
@lorenzo-stoakes Thanks for the good comments on there. Otherwise I'd probably forget what the heck some of that was doing in a month, haha. Merged! Thanks for digging into that issue. I like your fix.

lorenzo-stoakes:
Great, thanks a lot @benbjohnson ! It was a subtle and odd bug and I am glad to help contribute :)

PR #579, opened 2016-07-30
Update description of project using BoltDB

asdine:
The description was too old so i figured i could update it


benbjohnson:
👍 thanks @asdine!

PR #584, opened 2016-08-18
Fix Go 1.7 pointer reference bug

benbjohnson:
## Overview

This commit fixes a bug where page end-of-header pointers were being converted to byte slices even when the pointer did not point to allocated memory. This occurs with pages that have a `page.count` of zero.

Note: This was not an issue in Go 1.6 but the new Go 1.7 SSA backend handles `nil` checks differently.

See https://github.com/golang/go/issues/16772

/cc @zeebo

### DIFF
--- a/freelist.go
+++ b/freelist.go
@@ -166,12 +166,16 @@ func (f *freelist) read(p *page) {
 	}
 
 	// Copy the list of page ids from the freelist.
-	ids := ((*[maxAllocSize]pgid)(unsafe.Pointer(&p.ptr)))[idx:count]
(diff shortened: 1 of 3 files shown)


ebusto:
This fixes the issue for me. Thanks for the hard work on BoltDB and the quick fix for this issue.

xiang90:
@benbjohnson Can we do a release for this issue?

benbjohnson:
@xiang90 Released as v1.3.0: https://github.com/boltdb/bolt/releases/tag/v1.3.0

xiang90:
@benbjohnson Thanks a lot!

PR #590, opened 2016-09-05
Compaction Command

benbjohnson:
## Overview

This pull request combines the `bolt compact` changes from #460 with an additional commit with minor changes to coding style as well as the introduction of `Bucket.Sequence()` and `Bucket.SetSequence()`.


justinfx:
I was wondering if I could get some clarification on the use of Sequence and SetSequence when doing a compact operation? 

We have an application which uses `NextSequence` to get an ID which is then used as a key in the bucket. If we were to compact the database, does that mean the sequence is reset and we would need to call `SetSequence` with the last value of `Sequence`? If we used the compact command and failed to do this, would we end up with a sequence that starts over again and overwrites our existing keys?

justinfx:
Oh I see:
https://github.com/boltdb/bolt/pull/590/files#diff-106a9dcefc59220b0839abb6c8452fd1R1673

It uses the publicly exposed `Sequence()` and `SetSequence()` to preserve them in the destination database. And they are public so that users have that same access when walking manually.

PR #597, opened 2016-09-19
Improve node dumps

glevand:
Improve the node dumper by adding a verbosity value, sending the output to the
log, and always building the dumper code.

A verbosity of 2 gives the same output as before. Verbosity 3 gives a truncated
hexdump of the inodes.

Sending the output to the log is more flexible in that the log can be sent to a
file, which useful when the process is a spawned child.

By using conditions on constant variables the dumper code will be compiled to
catch build regressions.  The optimizer should remove the dead code when not
enabled.


glevand:
Rebased to latest.

PR #607, opened 2016-10-18
Add MAP_PRIVATE for mmap, if fs don't surport MAP_SHARED, we use MAP_PRIVATE instead

keloyang:
As we know, some fs don't surport MAP_SHARED, such as jffs2/logfs/9p/afs, and MAP_SHARED is not always needed, so we can use MAP_PRIVATE instead to increased usage scenario.
Signed-off-by: Shukui Yang <yangshukui@huawei.com>


benbjohnson:
I tried testing `MAP_PRIVATE` locally and tests fail immediately locally.

HuKeping:
Anymore details about the failure @benbjohnson ?

PR #613, opened 2016-10-28
Add Value compression

weberr13:
This is a possible method of adding snappy compression to bolt values.  Keys are unchanged and compression occurs on a per-page basis making pages with multiple similar values (json in our case) compress efficiently.  Performance is diminished primarily on the read side because whole pages must be reallocated instead of read directly from the mmfile.  This will be mitigated later with either a change to this PR or a new PR that uses a byte pool (not sync.Pool, probably https://github.com/vrecan/poolShark).  By default this is disabled and a configuration parameter is in place to enable.  Even when enabled it is backwards compatible for uncompressed.  In addition, there are several new Seek methods that skip value reading since it is redundant and actually matters in the case the values were compressed but unused.


benbjohnson:
@weberr13 Thanks for taking the time to add compression but I feel like this is too substantial of a change to Bolt and will negatively affect its stability. I can see this being useful either as a fork of Bolt or as a utility library separate from Bolt. For that reason I'm closing this PR.

PR #618, opened 2016-11-04
[ppc64] added missing variable

tbe:
The variable `brokenUnaligned` was missing for ppc64.

PR #622, opened 2016-11-14
Update bolt_ppc64.go

hashneo:
Missing variable on ppc64 arch. Not sure if it should be true or false?


benbjohnson:
Fixed in #618.

PR #625, opened 2016-11-16
replace unix implementation to be the same as solaris to fix an issue with glusterfs

### LINKED ISSUE #600: boltdb flock fails on glusterfs 3.7.9
I attempted to run boltdb from multiple nodes on [glusterfs](https://gluster.readthedocs.io/en/latest/) and it immediately deadlocks. When switching to [FcntlFlock](https://golang.org/pkg/syscall/#FcntlFlock) it behaves as you would expect.

In golang we have run into issues with flock in our software, mostly around it's interaction with [glusterfs](https://gluster.readthedocs.io/en/latest/). The solution we ended up sticking with was using [FcntlFlock](https://golang.org/pkg/syscall/#FcntlFlock) instead. It's compatable with all standard filesystems and behaves well with [glusterfs](https://gluster.readthedocs.io/en/latest/). 

I would like to put up a pr to fix this, would you prefer that I just replace the current implementation or would you want an option to enable  [FcntlFlock](https://golang.org/pkg/syscall/#FcntlFlock)?

vrecan:
fixes #600. For some reason this test fails for me locally. Ignore the previous comment that was because I was putting a timeout on my tests that was too low. All tests pass and glusterfs now works as expected when locking files.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -21,16 +21,21 @@ func flock(db *DB, mode os.FileMode, exclusive bool, timeout time.Duration) erro
 		} else if timeout > 0 && time.Since(t) > timeout {
 			return ErrTimeout
 		}
-		flag := syscall.LOCK_SH
+		var lock syscall.Flock_t
+		lock.Start = 0
+		lock.Len = 0
+		lock.Pid = 0
+		lock.Whence = 0
+		lock.Pid = 0
 		if exclusive {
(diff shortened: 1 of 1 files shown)


vrecan:
I've run on a 5 and 8 node glusterfs cluster with locking done on all nodes and things now behave correctly. Previously the system would deadlock within seconds.

PR #635, opened 2016-12-20
bucket: correct broken unaligned load/store in ppc64

michelmno:
similar patch as per ppc64le with cid 97aba55

Signed-off-by: Michel Normand <normand@linux.vnet.ibm.com>


benbjohnson:
Fixed in #618.

PR #641, opened 2016-12-23
Minor cleanup and bug fixes

josharian:
These all came from reading the code, although I believe the two bug fixes solve potentially real issues.


benbjohnson:
@josharian lgtm. Good catch on 7adfa44. That's definitely a bug. Cleaning up the `Tx` in 10c6e01 is great too.

PR #649, opened 2017-01-11
Prevent bus error caused by reading a corrupted database

gmlewis:
If a read-only database is corrupted and a new read is attempted, the `meta0` and/or `meta1` structs are likewise corrupted and calling `db.meta()` causes a `bus error`.

This patch calls `validate` on these structs before attempting to create a transaction which prevents the `bus error` entirely.

Change-Id: I798dac1a7cf36283fbc64ee63c3e2d80bca54e93

### DIFF
--- a/db.go
+++ b/db.go
@@ -481,6 +481,18 @@ func (db *DB) beginTx() (*Tx, error) {
 		return nil, ErrDatabaseNotOpen
 	}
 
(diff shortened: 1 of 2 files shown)


adriantam:
+1
Good fix

gmlewis:
friendly ping

Issue #653, opened 2017-01-30
frequent "misplaced bucket header" and "page already freed" errors


schollz:
Hi, I've been using bolt quite successfully for awhile for my project, [FIND](https://github.com/schollz/find).

Recent builds have been creating the following errors in one of my tests:

```
2017/01/30 16:47:32 [Recovery] panic recovered:
POST /track HTTP/1.1
Host: 127.0.0.1:8003
Accept: */*
Content-Length: 1089
Content-Type: application/json
Expect: 100-continue
User-Agent: curl/7.47.0


misplaced bucket header: 7265736f7572636573 -> 73766d44617461
/usr/local/go/src/runtime/panic.go:458 (0x47b6a3)
	gopanic: reflectcall(nil, unsafe.Pointer(d.fn), deferArgs(d), uint32(d.siz), uint32(d.siz))
/home/zns/work/src/github.com/boltdb/bolt/bucket.go:556 (0x5778c8)
	(*Bucket).spill: panic(fmt.Sprintf("misplaced bucket header: %x -> %x", []byte(name), k))
/home/zns/work/src/github.com/boltdb/bolt/tx.go:163 (0x584f55)
	(*Tx).Commit: if err := tx.root.spill(); err != nil {
/home/zns/work/src/github.com/boltdb/bolt/db.go:605 (0x57c96d)
	(*DB).Update: return t.Commit()
/home/zns/Documents/find/svm.go:124 (0x4429c9)
	dumpFingerprintsSVM: })
/home/zns/Documents/find/fingerprint.go:204 (0x4127c5)
	trackFingerprint: dumpFingerprintsSVM(group)
/home/zns/Documents/find/fingerprint.go:138 (0x41088a)
	trackFingerprintPOST: message, success, locationGuess, bayes, svm, rf := trackFingerprint(jsonFingerprint)
/home/zns/work/src/github.com/gin-gonic/gin/context.go:98 (0x58a81a)
	(*Context).Next: c.handlers[c.index](c)
/home/zns/work/src/github.com/gin-gonic/contrib/sessions/sessions.go:65 (0x62dbb9)
	Sessions.func1: c.Next()
/home/zns/work/src/github.com/gin-gonic/gin/context.go:98 (0x58a81a)
	(*Context).Next: c.handlers[c.index](c)
/home/zns/work/src/github.com/gin-gonic/gin/recovery.go:45 (0x59a8fa)
	RecoveryWithWriter.func1: c.Next()
[...]

PR #654, opened 2017-01-31
Revert "replace unix implementation to be the same as solaris to fix …"

benbjohnson:
Reverts a previous commit based on feedback from https://github.com/boltdb/bolt/issues/653#issuecomment-276383239.

/cc @schollz

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -21,21 +21,16 @@ func flock(db *DB, mode os.FileMode, exclusive bool, timeout time.Duration) erro
 		} else if timeout > 0 && time.Since(t) > timeout {
 			return ErrTimeout
 		}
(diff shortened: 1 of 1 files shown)


schollz:
Thanks, it seems to fix the problem. I will check the ARM systems later and let you know.

benbjohnson:
@schollz Thank you. Once you can test the ARM version then I'll merge it in.

schollz:
@benbjohnson Tested and it works well on ARM. Thanks again for the quick response and also bringing my attention to great utility of `git bisect`!

benbjohnson:
I am surprised that it's that commit that's causing the issue. Let me know if you have any further issues.

PR #679, opened 2017-05-01
Compact provides an API for regularly compressing the database.

### LINKED ISSUE #674: request: compaction api call (as bolt disk space consumption grows without bound)
I don't mind that a bolt file takes up 3-4x the space on disk as the contained value, but I had a bug where my usually 500KB blob ballooned to 7MB before I caught the leak.

Now my boltdb takes up 21MB instead of the 1.5MB, which costs alot as I ship the file around.

Feature request: an api call to compact an existing boltdb file so that it takes up only the pages that are actually in use.  I suppose I could just delete and re-write the file from scratch, but that introduces a bunch of race conditions that could have the file disappearing if the process is killed at the wrong moment. 

Would this be difficult to implement?

glycerine:
Without regular compaction, the bolt database will grow
without bound on disk.

Fixes #674

### DIFF
--- a/compact.go
+++ b/compact.go
@@ -1,0 +1,257 @@
+package bolt
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"sync/atomic"
+)
+
+// QuietlyCompact calls compact with the defaut batchSizeBytes
+// and does no logging of the impact of the compression.
+func (db *DB) CompactQuietly() error {
+	return db.Compact(0, false)
+}
+
+// Compact does these steps to reduce the
+// space fragmentation that happens in a bolt database.
+//
+// 0) Open a fresh .compact_in_progress bolt db file.
+// 1) Read each object from the bolt db and write it
+//    into the fresh .compact_in_progress database.
+// 2) Close the both files.
+// 3) Rename the .compact_in_progress file to be the original db file name.
+//    This leverages the fact that os.Rename is atomic.
+// 4) Re-open the newly compact-ed db file.
+//
+// INVAR: db must be already open.
+//
+// Side effect:
+//  The file db.path+".compact_in_progress" will be
+//  deleted at the top of Compact(), then used to
+//  create the newly compacted database.
+//
+// batchSizeInBytes should supply the threshold size in bytes
+// after which we commit a transaction. The default of 4MB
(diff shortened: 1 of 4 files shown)


glycerine:
@benbjohnson if you are too busy to review, is there somebody else you could suggest?

vincent-petithory:
Most of the code seems to come from the existing  compact cmd, so it should
be fine.

Le lun. 1 mai 2017 18:26, Jason E. Aten, Ph.D. <notifications@github.com> a
écrit :

benbjohnson:
@glycerine I'm hesitant to add this feature because I think it adds a lot of code and complexity for an issue that relatively few people have. Bolt doesn't typically work well with large values since it causes fragmentation.

This is also a feature that can largely be implemented outside of the core Bolt codebase. Why not make this a separate library that is layered on?

glycerine:
This allows bolt to work with large values, because it allows the fragmentation they produce to be addressed quite effectively. I run regularly with a single value, 3MB blob.

We must close the old database file and re-open the newly compacted file, so all places where the database is in use must be told to stop and start using a new database handle. This is difficult and awkward, if it is possible at all, from the outside library. By making the change inside the DB structure itself, everyone starts using the newly compacted database in unison at once.

benbjohnson:
@glycerine I thought it over for day and I decided that it's too much of an edge case feature to warrant the large amount of code required. I think the best case forward would be to leave it as a branch on your fork and we can add a link in the README for anyone who needs the functionality. It's easy enough to vendor in a different repo in a project and the `bolt` project doesn't change often enough that the fork would get out of date.

adamluzsi:
Still the concept is awesome and as a package that wraps boltdb connection around and add this feature is a promising idea.

glycerine:
@benbjohnson Thanks for thinking about it. I guess I just disagree about the importance of being able to store large values. But I understand that you want to keep the codebase small, and that is a valuable attribute.

@adamluzsi I don't see how to make a wrapper, since we need to add a check at the end of every transaction commit. If there was a hook for all transactions (not just on a tx-by-tx basis), then a wrapper might be doable.

I've made a fork to support big values, https://github.com/bigboltdb/bolt.

PR #680, opened 2017-05-03
provide CommitHook callback, to hook all commits.

glycerine:
Hooking all commits could allow us to implement wrappers.

For example, this would support a wrapper approach to compaction rather than the https://github.com/bigboltdb/bolt fork.


glycerine:
I'm going to get more experience with this, then resubmit once I've got a wrapper lib working-- to make sure the hooks are sufficient before merging upstream.

PR #682, opened 2017-05-03
avoid data race when reading from read-only tx

heyitsanthony:
Concurrently getting buckets and creating cursors on a read-only tx will
trigger a data race on the cursor counter and the bucket subcache. Instead,
use atomics and locks to protect the shared state.

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -3,6 +3,8 @@ package bolt
 import (
 	"bytes"
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"unsafe"
 )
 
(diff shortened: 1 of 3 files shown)


xiang90:
https://github.com/boltdb/bolt#transactions says

" Individual transactions and all objects created from them (e.g. buckets, keys) are not thread safe.
 To work with data in multiple goroutines you must start a transaction for each one or use locking to 
ensure only one goroutine accesses a transaction at a time. Creating transaction from the DB is thread safe. "

I am not sure if this PR can make Read Only transaction 100% thread safe.

PR #694, opened 2017-06-05
Garbage collect pages allocated after minimum txid

heyitsanthony:
Read txns would lock pages allocated after the txn, keeping those pages
off the free list until closing the read txn. Instead, track allocating
txid to compute page lifetime, freeing pages if all txns between
page allocation and page free are closed.

For etcd, this reduces a free list due to read txn locking from 8k pages to about 3k. It can get down to about 800 pages, but some pages are being allocated outside the freelist alloc path and it's not obvious to me where and how to assign the allocating txid.

/cc @xiang90

### DIFF
--- a/freelist.go
+++ b/freelist.go
@@ -6,18 +6,29 @@ import (
 	"unsafe"
 )
 
+
+// txPending holds a list of pgids and corresponding allocation txns
+// that are pending to be freed.
+type txPending struct {
+	ids              []pgid
+	alloctx          []txid // txids allocating the ids
+	lastReleaseBegin txid   // beginning txid of last matching releaseRange
+}
+
 // freelist represents a list of all pages that are available for allocation.
 // It also tracks pages that have been freed but are still in use by open transactions.
 type freelist struct {
-	ids     []pgid          // all free and available free page ids.
-	pending map[txid][]pgid // mapping of soon-to-be free page ids by tx.
-	cache   map[pgid]bool   // fast lookup of all free and pending page ids.
+	ids     []pgid              // all free and available free page ids.
+	allocs  map[pgid]txid       // mapping of txid that allocated a pgid.
+	pending map[txid]*txPending // mapping of soon-to-be free page ids by tx.
+	cache   map[pgid]bool       // fast lookup of all free and pending page ids.
 }
 
 // newFreelist returns an empty, initialized freelist.
 func newFreelist() *freelist {
 	return &freelist{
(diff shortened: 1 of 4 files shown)


freelist.go
```diff
@@ -6,18 +6,29 @@ import (
 	"unsafe"
 )
 
+
+// txPending holds a list of pgids and corresponding allocation txns
+// that are pending to be freed.
+type txPending struct {
+	ids              []pgid
+	alloctx          []txid // txids allocating the ids
```
xiang90: why do we need to record alloctx. i thought the pages will only be freed/allocated by RW tx. So when we decide to free it up, the allocator tx should be closed (allocated by a previous rw tx) already or closing (allocated by this tx).

heyitsanthony: 1, open read txns txid=7,8,9
2. allocate p at txid=10; close txn
3. open read txns txid=11,12; both depend on p
4. free p at txid=13; close txn

Closing txid=13 can't free p until 11 and 12 are closed, but it doesn't have to wait for 7,8,9 to close since the minimum txn (tracked by alloctx) that can possibly use the page is txid=10.

xiang90: make sense.

xiang90:
I took a quick look. It makes sense to me.

@benbjohnson do you have any time to take a look? Or do you have any suggestions to how to get this merged? Eventually, we want to be able to free ALL possible pages even if there is a long running read tx. But it might take us multiple steps to get there.

One more thing we might want to do is not to record freelist on disk at all. We will recompute it for every startup by scanning the entire db. In our use case, the DB size is small, and we do want to save the io to write down freelists when there are quite some free pages.

rgeronimi:
If such a « no persistence » mode for freelist is merged, it needs to be protected by a configuration flag in Options. We run a database with more than a hundred gigabytes, with very short program startup time needed (less than a few seconds), so we definitely can’t afford a rescan upon restart.

Regards,

Raphael Geronimi

xiang90:
> If such a « no persistence » mode for freelist is merged, it needs to be protected by a configuration flag in Options. 

Yes. It will be an option if we decide to implement it.

xiang90:
> We run a database with more than a hundred gigabytes, with very short program startup time needed 

One off of topic question: have you measured your freelist size when the db grows to 100s GB?

rgeronimi:
No

Most of the content is dormant. We sometimes add documents, but we rarely update or delete them.

We read a lot.

Raphael

PR #695, opened 2017-06-08
add option to skip freelist sync

xiang90:
When the database has a lot of freepages, the cost to sync all
freepages down to disk is high. If the total database size is
small (<10GB), and the application can tolerate ~10 seconds
recovery time, then it is reasonable to simply not sync freelist
and rescan the db to rebuild freelist on recovery.

This is a early attempt. Would like to see your feedbacks.

/cc @benbjohnson @heyitsanthony


xiang90:
In our test environment, this reduces the db size from ~100MB to ~10MB and reduce 5x iops.

lrita:
In my opinion, this is a good idea to reduce IOPS. But, this implement maybe cause database corruption for now, if there are some k-v pairs' length large than a pagesize. `page. overflow` is relation to this problem.

```
type page struct {
	id       pgid
	flags    uint16
	count    uint16
	overflow uint32
	ptr      uintptr
}
```

xiang90:
@lrita fixed.

xiang90:
talked with @benbjohnson and confirmed this will not get merged. so closing.

PR #696, opened 2017-06-11
Set FillPercent=1.0 in 'bolt compact'.

dogben:
By default, pages are split when they reach half full. For 'bolt compact' we want to fill the entire page for maximum compaction.

I tested on a copy of a production DB of 2.1G. Without setting FillPercent, 'bolt compact' outputs a 1.6G file. After this change, 'bolt compact' outputs a 1.1G file.


dogben:
@benbjohnson @vincent-petithory

PR #697, opened 2017-06-15
change to normal instead of random for mmap madvise

### LINKED ISSUE #691: high iops because of MADV_RANDOM when doing sequential access
I noticed that on high latency filesystems (Elastic file system, GlusterFS) we were seeing really terrible performance with boltdb. I dug in a little bit more and noticed that we were advising the kernel that we are doing random IO. In my case this is not true, almost all of our io is sequential. 

Original code
```
	if err := madvise(b, syscall.MADV_RANDOM); err != nil {
		return fmt.Errorf("madvise: %s", err)
	}
```
changing to this
```
	if err := madvise(b, syscall.MADV_NORMAL); err != nil {
		return fmt.Errorf("madvise: %s", err)
	}
```

We got more then a 60x performance increase. We were seeing reads of around 1500kb/s (~500iops) before and that changed to 60000kb/s (~100iops). We also saw a massive decrease in the number of iops required to read from the db.  I was thinking about putting up a pr that allowed you to choose which MADV flag we use inside of boltdb. Does this sound like something you would accept?

vrecan:
This fixes #691, I can adjust this to be more configurable but this is how I fixed my issue.

### DIFF
--- a/bolt_unix.go
+++ b/bolt_unix.go
@@ -53,7 +53,7 @@ func mmap(db *DB, sz int) error {
 	}
 
 	// Advise the kernel that the mmap is accessed randomly.
-	if err := madvise(b, syscall.MADV_RANDOM); err != nil {
+	if err := madvise(b, syscall.MADV_NORMAL); err != nil {
 		return fmt.Errorf("madvise: %s", err)
 	}
 


funny-falcon:
But it was changed to RANDOM because in other usage patterns it is faster.

Looks like it should be configuration option.

PR #700, opened 2017-06-18
add option to skip freelist sync

xiang90:
When the database has a lot of freepages, the cost to sync all
freepages down to disk is high. If the total database size is
small (<10GB), and the application can tolerate ~10 seconds
recovery time, then it is reasonable to simply not sync freelist
and rescan the db to rebuild freelist on recovery.


xiang90:
wrong branch... closing.

PR #701, opened 2017-06-20
*: use coreos/bbolt as import path

xiang90:


xiang90:
sorry... wrong again. wont do this again.

PR #714, opened 2017-08-05
Added etcd to projects using boltdb

dankohn:
There's a good description of how boltdb is used in etcd at https://coreos.com/blog/performance-of-etcd.html


heyitsanthony:
@dankohn etcd doesn't use upstream boltdb anymore; we're maintaining an active fork at https://github.com/coreos/bbolt with new performance improvements and bug fixes

dankohn:
@heyitsanthony Good to know, thanks.

PR #402, opened 2015-07-28
Add "buckets", "keys" and "get" CLI commands

lmars:
These were previously removed in https://github.com/boltdb/bolt/commit/d0e8a99e30eebd30669908bbe57c969e8921ca6c but I find them quite useful so have reintroduced them.

Closes https://github.com/boltdb/bolt/issues/400.

PR #715, opened 2017-08-11
Add "buckets", "keys" and "get" CLI commands

heyitsanthony:
Looks reasonable.

via https://github.com/boltdb/bolt/pull/402


heyitsanthony:
oops

PR #721, opened 2017-08-21
Add NoSync field to Options

tylertreat:
Proposed change to expose `NoSync` in `Options`. Hopefully this is a reasonable change. The justification for this is I use a number of libraries which simply expose the `Options` passed into `Open()` but not the underlying `DB` they create. To enable `NoSync` in these cases, I either have to fork the library or fork bolt.


heyitsanthony:
@tylertreat is switching to a fork feasible for your use case? if so this patch could go into the fork at http://github.com/coreos/bbolt which includes some extra enhancements/bugfixes

tylertreat:
@heyitsanthony is there more information on the fork somewhere? It looks like the README doesn't detail the purpose/goal of the fork.

heyitsanthony:
@tylertreat the purpose is to actively maintain boltdb; this project is being ignored.

tylertreat:
@heyitsanthony I mirrored the PR here: https://github.com/coreos/bbolt/pull/30

PR #740, opened 2018-01-18
Make boltdb build on ppc

mvo5:
Trivial PR to fix two problems on ppc:
1. without the `var brokenUnaligned` the build fails with an undefined var
1. the mmap size does not fit into a 32bit register


benbjohnson:
Hi @mvo5. This repo is not accepting pull requests anymore. Looks like your pull request was accepted by the CoreOS fork.

